PR #20, opened 2016-09-28
added debug screen

amulware:
@willsb please review, kthxbye


teferi/ViewControllers/MainViewController.swift
```diff
@@ -23,13 +23,17 @@ class MainViewController : UIViewController, MFMailComposeViewControllerDelegate
     }
     
     @IBOutlet private weak var titleLabel : UILabel!
+    @IBOutlet weak var debugView: DebugView!
```
willsb: y u no private?

teferi/Views/TimelineCell.xib
```diff
@@ -1,19 +1,20 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
willsb: These changes only make merging harder, it seems

Issue #8, opened 2016-09-02
Metrics needed


pe0ny:
- [x] Number of downloads through lifetime of the app (that we can get from itunes connect)
- [x] Number of uninstalls through lifetime of the app
- [x] Number of unique sessions (the number of times the app has been used for at least two seconds)
- [x] Number of total sessions (the number of times the app has been used for at least two seconds by all users) daily/weekly/monthly
- [x] Time spent per session to measure user engagement of the app (similar to bounce rate)
- [x] Conversion rate = app openings/total downloads
- [x] Rating

pe0ny:
And we want to measure usage of:
- [ ] Edit function
- [ ] Switch function ('+' button)

willsb:
PR #12 adds most of the needed functionality. We can use HockeyApp's custom events to add the Edit/Create button events, once we have the proper UI for them. I just need to figure out if Hockey App gives us the time spent on each session.

amulware:
Blocked by #7

PR #26, opened 2016-09-30
Added hockeyapp custom events

amulware:
See #8


teferi/Services/HockeyAppMetricsService.swift
```diff
@@ -15,8 +15,9 @@ class HockeyAppMetricsService : MetricsService
     }
     
     ///Used to send custom events to the framework
-    func log(customEvent event: CustomEvent)
+    func log(event: CustomEvent)
     {
-        //TODO: Add loggin of custom events once their UI is added
+        let metricsManager = BITHockeyManager.shared().metricsManager
+        metricsManager.trackEvent(withName: "\(event)")
```
willsb: I believe it's more correct to `event.rawValue` instead

teferi/ViewControllers/TimelineViewController.swift
```diff
@@ -21,13 +21,17 @@ class TimelineViewController : UITableViewController
     
     init(date: Date)
     {
-        viewModel = TimelineViewModel(date: date, persistencyService: self.appDelegate.persistencyService)
+        viewModel = TimelineViewModel(date: date,
+                                      persistencyService: self.appDelegate.persistencyService,
+                                      metricsService: self.appDelegate.metricsService)
```
willsb: You'll have to rewrite the tests because of this change

PR #37, opened 2016-10-10
Replaces HockeyApp with Fabric

### LINKED ISSUE #8: Metrics needed

willsb:
Closes #8


willsb:
cc @amulware

PR #72, opened 2016-10-19
Permissions overlay

### LINKED ISSUE #36: Empty(?) state for disabled location permissions
As mentioned in #16, the app should show a friendly message in the main view if location permissions are disabled, directing the user to enable them in the settings.

This needs design/copy.

I think we should still allow the user to use the app manually in the meantime however, and not shut people out completely, if they have locations disabled.

willsb:
Closes #36


amulware:
Did a quick test, and at first it showed me the overlay because location was disabled for the app. Then I enabled it, went back to the app and the overlay was gone. I then closed the app, and disabled location in the settings again. Going back to the app: no overlay.

amulware:
I suppose it did not ask me again because of the 24 hour limit?

amulware:
See comment in #36 for more requirements.

PR #92, opened 2016-10-24
Fixes location service tracking

### LINKED ISSUE #87: Fix location tracking
It currently (0.5) seems that the location algorithm is not creating any time slots. This could be because the algorithm is broken, or because it does not receive any location updates.

willsb:
Closes #87


amulware:
Will leave this here until we performed a live test.

PR #94, opened 2016-10-24
Fixes Black Screen glitch

### LINKED ISSUE #88: Find and resolve black-screen issue
When leaving the app running over night, and sometimes when opening it from a notification, it shows up only as a black screen and needs to be restarted.

willsb:
Closes #88


teferi/AppDelegate.swift
```diff
@@ -62,6 +62,15 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         //Start metrics
         self.metricsService.initialize()
```
amulware: I believe this line should also be in `initializeInterfaceIfNeeded()`.

PR #104, opened 2016-10-26
Fix onboarding permissions refresh

askatasuna:


teferi/Onboarding/OnboardingPage.swift
```diff
@@ -10,20 +10,30 @@ class OnboardingPage : UIViewController
...
-    private var onboardingPageViewController : OnboardingPageViewController!
+    private(set) var onboardingPageViewController : OnboardingPageViewController!
     
     init?(coder aDecoder: NSCoder, nextButtonText: String?)
     {
         super.init(coder: aDecoder)
         
         self.nextButtonText = nextButtonText
+        
+        NotificationCenter.default.addObserver(self,
```
willsb: Why subscribe globally instead of using the `AppStateService`?

PR #120, opened 2016-11-02
iOS 9.3 edit hotfix

### LINKED ISSUE #119: Fix editing not working on iOS 9.3

willsb:
Closes #119


teferi/Views/TimelineCell.swift
```diff
@@ -12,32 +12,41 @@ class TimelineCell : UITableViewCell
     private let minuteMask = "%02d min"
     private lazy var lineHeightConstraint : NSLayoutConstraint =
     {
-        return NSLayoutConstraint(item: self.lineView!, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1.0, constant: CGFloat(Constants.minLineSize))
+        for constraint in self.lineView.constraints
+        {
+            guard constraint.firstAttribute == .height else { continue }
+            return constraint
```
amulware: this loop body should really be a clear single line `if` statement

teferi/Views/TimelineCell.swift
```diff
@@ -12,32 +12,41 @@ class TimelineCell : UITableViewCell
...
     {
-        //[aView convertPoint:localPosition toView:nil];
         self.isSubscribedToClickObservable = true
-        return self.categoryButton!.rx.tap
-            .map { return (self.categoryIcon!.convert(self.categoryIcon!.center, to: nil), self.currentIndex) }
+        
+        return self.categoryButton.rx.tap
+            .map
+            {
+                return (self.categoryIcon.convert(self.categoryIcon!.center, to: nil), self.currentIndex)
```
amulware: I believe this line is wrong (and was so before) even though it happens to work. I believe the first parameter should be half the size of the category icon, and not the center(which is relative to the icon's parent). Or is there a way why that difference is important?

teferi/Views/TimelineCell.swift
willsb: Nesting is gross, but I can change if you really want

teferi/Views/TimelineCell.swift
willsb: `self.categoryIcon.convert(_, to)` finds the center of the view and returns the position relative to the window and not to the parent

teferi/Views/TimelineCell.swift
amulware: I would appreciate that, I think it is clearer. After all, the `continue` is also nested.

teferi/Views/TimelineCell.swift
amulware: The center of the view is not the same as half the view's size though. The center is instead relative to the view's parent's coordinates. If that is intended here, then ignore this.

PR #131, opened 2016-11-10
Feature/actionable notifications ios10

sofokli1:
Quick way to test new functionality please remove `private` from `AppDelegate line 21` and add the following snippet to `TimelineViewModel line 84`
`let notificationDate = Date().addingTimeInterval(5)
let notificationService = ((UIApplication.shared.delegate as? AppDelegate)?.notificationService)!
notificationService.scheduleNotification(date: notificationDate, title: "Howdy!", message: "What you’re doing right now?")`

### DIFF
--- a/NotificationContentExtension/NotificationViewController.swift
+++ b/NotificationContentExtension/NotificationViewController.swift
@@ -1,0 +1,51 @@
+//
+//  NotificationViewController.swift
+//  NotificationContentExtension
+//
+//  Created by Juxhin Bakalli on 08/11/2016.
+//  Copyright © 2016 Toggl. All rights reserved.
+//
+
+import UIKit
+import UserNotifications
+import UserNotificationsUI
+
+class NotificationViewController: UIViewController, UNNotificationContentExtension {
+
+    // MARK: - Properties
+    @IBOutlet weak var stackView: UIStackView!
+    
+    func didReceive(_ notification: UNNotification) {
+
+        if let timeSlots = notification.request.content.userInfo["timeSlots"] as? [[String: String]] {
+            
+            for timeSlot in timeSlots {
+                var color = UIColor.gray
+                var category: String?
+                var date = String()
+                
+                if let colorHex = timeSlot["color"] {
+                    color = UIColor(hexString: colorHex)
+                }
+                
+                if let categoryString = timeSlot["category"] {
+                    category = categoryString
+                }
+                
+                if let dateString = timeSlot["date"] {
+                    date = dateString
+                }
+                
+                let timeSlotView = TimeSlotMiniView.instanceFromNib()
+                timeSlotView.setUI(color, category: category, date: date)
+                stackView.addArrangedSubview(timeSlotView)
+            }
+            
+            if let lastSlotView = stackView.arrangedSubviews.last as? TimeSlotMiniView {
+                UIView.animate(withDuration: 0.5, delay: 0, options: [.autoreverse, .repeat], animations: {
+                    lastSlotView.colorView.transform = CGAffineTransform(scaleX: 1.4, y: 1.4)
+                }, completion: nil)
+            }
+        }
+    }
+}
--- a/NotificationContentExtension/NotificationViewController.swift
+++ b/NotificationContentExtension/NotificationViewController.swift
@@ -1,11 +1,3 @@
-//
-//  NotificationViewController.swift
-//  NotificationContentExtension
-//
-//  Created by Juxhin Bakalli on 08/11/2016.
-//  Copyright © 2016 Toggl. All rights reserved.
-//
-
 import UIKit
 import UserNotifications
 import UserNotificationsUI
--- a/teferi.xcodeproj/project.pbxproj
+++ b/teferi.xcodeproj/project.pbxproj
@@ -86,6 +86,7 @@
 		512A9B3D1DD4450200B73F01 /* TimeSlotMiniView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 512A9B3C1DD4450200B73F01 /* TimeSlotMiniView.swift */; };
 		512A9B3F1DD4451F00B73F01 /* TimeSlotMiniView.xib in Resources */ = {isa = PBXBuildFile; fileRef = 512A9B3E1DD4451F00B73F01 /* TimeSlotMiniView.xib */; };
 		512A9B411DD46E3B00B73F01 /* ColorExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 24564F611D83265900D9A8AE /* ColorExtensions.swift */; };
+		512A9B431DD49AD200B73F01 /* TimeSlotChangeType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 512A9B421DD49AD200B73F01 /* TimeSlotChangeType.swift */; };
 		513535D31DA26E3B00A365A9 /* Onboarding.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 513535D11DA26E3B00A365A9 /* Onboarding.storyboard */; };
 		513535D61DA2756C00A365A9 /* OnboardingPageViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 513535D51DA2756C00A365A9 /* OnboardingPageViewController.swift */; };
 		51BED8F71DD26FA800455EF6 /* UserNotifications.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 51BED8F61DD26FA800455EF6 /* UserNotifications.framework */; };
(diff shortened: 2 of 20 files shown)


teferi.xcodeproj/project.pbxproj
```diff
@@ -1153,12 +1281,12 @@
...
 				CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = NO;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
-				DEVELOPMENT_TEAM = B227VTMZ94;
+				DEVELOPMENT_TEAM = QK3BJ9S38G;
 				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
 				INFOPLIST_FILE = teferi/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				OTHER_SWIFT_FLAGS = "$(inherited) \"-D\" \"COCOAPODS\" -DDEBUG";
-				PRODUCT_BUNDLE_IDENTIFIER = com.toggl.superday;
+				PRODUCT_BUNDLE_IDENTIFIER = com.toggl.superday.sofo;
```
amulware: There are a bunch of changes to the development team, and bundle identifier all through this file. I'm not 100% sure this will harm us, but it seems risky.

teferi/AppDelegate.swift
```diff
@@ -66,6 +67,10 @@ class AppDelegate : UIResponder, UIApplicationDelegate
             return true
         }
         
+        if #available(iOS 10.0, *) {
+            (self.notificationService as? PostiOSTenNotificationService)?.setUserNotificationActions()
+        }
+
```
amulware: This is only run when the app starts in the foreground, and will not be run if the app starts in the background, or gets activated by the user after it started in the background. Should that code be here? Does it not need to cover the other cases?

teferi/Models/Category.swift
```diff
@@ -27,17 +27,37 @@ enum Category : String
...
         case .commute:
-            return UIColor(hexString: "#63D5EE")
+            return UIColor(hexString: self.colorHex)
         case .food:
-            return UIColor(hexString: "#FF6453")
+            return UIColor(hexString: self.colorHex)
         case .unknown:
-            return UIColor(hexString: "#CECDCD")
+            return UIColor(hexString: self.colorHex)
+        }
```
amulware: This switch-case is a bit silly now, isn't it? :D

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -46,20 +46,24 @@ class CoreDataPersistencyService : PersistencyService
...
+            fetchRequest.fetchLimit = numberOfTimeSlots
+        }
+        fetchRequest.sortDescriptors = [NSSortDescriptor(key: "startTime", ascending: false)]
         
         do
         {
             let results = try getManagedObjectContext().fetch(fetchRequest) as! [NSManagedObject]
             
-            let timeSlots = results.map(mapManagedObjectToTimeSlot)
+            let timeSlots = results.reversed().map(mapManagedObjectToTimeSlot)
```
amulware: Why is this reversing the results? Won't this invert the order of the timeline?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -99,6 +103,7 @@ class CoreDataPersistencyService : PersistencyService
             managedTimeSlot.setValue(category.rawValue, forKey: "category")
             
             try managedContext.save()
+            callbacks.forEach { callback in callback(timeSlot, .update) }
```
amulware: What is this change necessary for?

teferi/Services/PostiOSTenNotificationService.swift
```diff
@@ -44,4 +63,58 @@ class PostiOSTenNotificationService : NotificationService
         
         notifications.forEach { n in UIApplication.shared.cancelLocalNotification(n)  }
     }
+    
+    func handleNotificationAction(withIdentifier identifier: String?, for notification: UILocalNotification, completionHandler: @escaping () -> Void) {
+        if let identifier = identifier {
+            
+            var category = Category.commute
+            
+            switch identifier {
```
amulware: Rather than this conversion using constants, can't we just use the .rawValue of the category enum?

teferi/ViewControllers/TimelineViewController.swift
```diff
@@ -67,19 +67,40 @@ class TimelineViewController : UITableViewController
...
+        case .none:
+            
+            self.tableView.reloadData()
+            
+            let scrollIndexPath = IndexPath(row: viewModel.timeSlots.count, section: 0)
+            self.tableView.scrollToRow(at: scrollIndexPath, at: .bottom, animated: true)
+            
+        default:
+            break
+        }
```
amulware: These are big changes, can you give a quick summary of what exactly they do?

teferi/ViewModels/TimelineViewModel.swift
```diff
@@ -41,31 +49,48 @@ class TimelineViewModel
...
-        //Finishes last task, if needed
-        if let lastTimeSlot = timeSlots.last
-        {
-            lastTimeSlot.endTime = Date()
+        switch changeType {
+        case .create:
+            
+            if let lastTimeSlot = timeSlots.last
+            {
+                lastTimeSlot.endTime = Date()
```
amulware: This looks like a hack. Not important for this PR, but should be looked at. @willsb

teferi/ViewModels/TimelineViewModel.swift
```diff
@@ -41,31 +49,48 @@ class TimelineViewModel
         self.date = date.ignoreTimeComponents()
         self.persistencyService = persistencyService
         self.timeSlotsVariable = Variable(timeSlotsForDate)
+        self.timeSlotChangeVariable = Variable(TimeSlotChangeType.none)
         self.isEditingObservable = self.isEditingVariable.asObservable()
         self.timeSlotsObservable = self.timeSlotsVariable.asObservable()
+        self.timeSlotChangeObservable = self.timeSlotChangeVariable.asObservable()
```
amulware: What are the new Variable and Observable used for?

NotificationContentExtension/NotificationViewController.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  NotificationViewController.swift
+//  NotificationContentExtension
+//
+//  Created by Juxhin Bakalli on 08/11/2016.
+//  Copyright © 2016 Toggl. All rights reserved.
+//
```
willsb: //TODO: Add this to the Readme.md guidelines

teferi.xcodeproj/project.pbxproj
willsb: These changes should not be commited, as he just used them because of the Provisioning Profiles

sofokli1: Will fix. I have worked with `fastlane` tool `match` with one shared account between developers. It did wonders regarding this kind of issues + only one user needed to know the account password (for setup reasons). It abstracts the signing and certificate creation process [https://codesigning.guide](url) Just a thought.

teferi/Services/CoreDataPersistencyService.swift
sofokli1: Before modifying this function the order of the data was ascending. I modified the function so that it loads a specific number of timeSlots (or all) for a certain day. Since it requires to return the last N timeSlots I added a sortDescriptor to the fetchRequest in order to return the data in descending order in relation to the startTime (so that we get the latest data first). Afterwords the result from the fetchRequest is reversed so that the function returns data in ascending order as it was before modification it and as it is needed for the actionable notifications. We could add a `ascending: Bool` as a parameter to the function so that is more clear how the data are returned. What do you think @amulware ?

teferi/AppDelegate.swift
sofokli1: Based on the documentations there isn't much need to call this function again somewhere else.

> This method registers all of your categories at once, replacing any previously registered categories with the new ones in the categories parameter. Typically, you call this method only once at launch time.

teferi/Services/CoreDataPersistencyService.swift
sofokli1: It wasn't clear what kind of changes were happening to timeSlot so the `TimelineViewModel` was handling all of them as new timeSlot. This was needed because the changes that were done on the timeSlot based on the notification action were not propagated to the UI.

teferi/Services/PostiOSTenNotificationService.swift
sofokli1: That sound better. Will fix

teferi/ViewControllers/TimelineViewController.swift
sofokli1: The function was only supporting the case were new timeSlot where added to the `[TimeSlots]`. When notification changed the last time slot it could perform the needed UI updates. So instead of monitoring the array it observes for `TimeSlotChangeType` so that it is clear what happened to the array and what kind of action it needs to perform.

teferi/ViewModels/TimelineViewModel.swift
sofokli1: Shows the kind of changes that have happened to the `timeSlots'

teferi/Services/CoreDataPersistencyService.swift
amulware: No need to add a parameter we don't need right now.

teferi/AppDelegate.swift
sofokli1: For the app to run in the background or any other way it needs to at least have been opened once in the foreground, which is enough for the action registration to take place. Even if the user starts the app for the first time and does not allow notifications, the notification action will  still register. I can't think of a situation where the registration does not happen at least once (which is all we need).

amulware:
@willsb I'll let you have the final word on this, since it's your branch and you started work on this originally.

PR #135, opened 2016-11-21
Service Refactor

willsb:


teferi/Extensions/PredicateExtensions.swift
```diff
@@ -0,0 +1,10 @@
+import CoreData
+
+extension Predicate
+{
+    func convertToNSPredicate() -> NSPredicate
```
amulware: Why is this here and not just in the Predicate class?

teferi/Models/Predicate.swift
```diff
@@ -0,0 +1,11 @@
+class Predicate
```
amulware: Why use this class and not NSPredicate directly?

teferi/Models/BaseModel.swift
```diff
@@ -0,0 +1,15 @@
...
+import CoreData
+
+class BaseModel : NSObject
+{
+    required override init()
+    {
+        
+    }
+    
+    func setFromManagedObject(managedObject: NSManagedObject)
```
amulware: Why does the model know how to set itself from a managed object, but not the other way around?

amulware: Also the parameter should probably be unnamed.

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -16,52 +14,28 @@ class CoreDataPersistencyService : PersistencyService
...
-        {
-            self.loggingService.log(withLogLevel: .error, message: "Failed to create new TimeSlot")
-            return false
-        }
+        let element = self.mapManagedObjectIntoElement(managedElement)
+        return element
     }
     
-    func getTimeSlots(forDay date: Date) -> [TimeSlot]
+    override func get(predicate: Predicate) -> [ T ]
```
amulware: The parameter should maybe be exposed as `withPredicate`/`fromPredicate` or similar.

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -16,52 +14,28 @@ class CoreDataPersistencyService : PersistencyService
...
         
         do
         {
             let results = try self.getManagedObjectContext().fetch(fetchRequest) as! [NSManagedObject]
             
-            let timeSlots = results.map(self.mapManagedObjectToTimeSlot)
-            self.loggingService.log(withLogLevel: .info, message: "\(timeSlots.count) TimeSlots found")
-            return timeSlots
+            let elements = results.map(self.mapManagedObjectIntoElement)
+            loggingService.log(withLogLevel: .info, message: "\(elements.count) TimeSlots found")
```
amulware: "TimeSlots" in generic service.

amulware: Also, why is this logged in the first place, when other methods in this class are not?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -71,79 +45,108 @@ class CoreDataPersistencyService : PersistencyService
...
+            return true
+        }
+        catch
+        {
+            return false
+        }
     }
     
-    @discardableResult func updateTimeSlot(_ timeSlot: TimeSlot, withCategory category: Category) -> Bool
+    override func update(_ predicate: Predicate, updateFunction: (AnyObject) -> ()) -> Bool
```
amulware: Here the predicate parameter is unnamed. This should probably be consistent.

amulware: Further, why is the update function taking an `AnyObject` parameter? The way it is used is that it exposes a managed object, which is not clear here. Further, exposing a managed object seems counter the point of this service. Should it instead not give a `T` to update?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -71,79 +45,108 @@ class CoreDataPersistencyService : PersistencyService
...
-    
     //MARK: Methods
     private func getManagedObjectContext() -> NSManagedObjectContext
     {
         let appDelegate = UIApplication.shared.delegate as! AppDelegate
         return appDelegate.managedObjectContext
     }
     
-    private func mapManagedObjectToTimeSlot(_ managedObject: NSManagedObject) -> TimeSlot
+    private func setManagedProperties(_ managedElement: NSManagedObject, _ element: T)
```
amulware: See above. Why do we use reflection here, while the other way around the model takes care of it?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -71,79 +45,108 @@ class CoreDataPersistencyService : PersistencyService
...
     
-    private func getLastManagedTimeSlot() -> AnyObject?
+    private func getLastManagedElement() -> NSManagedObject?
     {
         let managedContext = self.getManagedObjectContext()
         
         let request = NSFetchRequest<NSFetchRequestResult>()
-        request.entity = NSEntityDescription.entity(forEntityName: timeSlotEntityName, in: managedContext)!
+        request.entity = NSEntityDescription.entity(forEntityName: self.getEntityName(), in: managedContext)!
         request.fetchLimit = 5
```
amulware: Why does it fetch five when it only returns one?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -152,38 +155,17 @@ class CoreDataPersistencyService : PersistencyService
         }
     }
     
-    private func endPreviousTimeSlot(atDate date: Date) -> Bool
+    private func getEntityName() -> String
```
amulware: This is used throughout and should be cached. Alternatively - especially if the model takes care of it's own conversion to/from managed objects - this could be moved to a static property in the model which could simply return the string.

teferi/Services/DefaultTimeSlotService.swift
```diff
@@ -0,0 +1,100 @@
...
+        
+        if !self.persistencyService.update(predicate, updateFunction: editFunction)
+        {
+            self.loggingService.log(withLogLevel: .error, message: "Error updating category of TimeSlot created on \(timeSlot.startTime) from \(timeSlot.category) to \(category)")
+        }
+    }
+    
+    func getLast() -> TimeSlot
+    {
+        return self.persistencyService.getLast() ?? TimeSlot()
```
amulware: Should this return a new time slot if none is found?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -71,79 +45,108 @@ class CoreDataPersistencyService : PersistencyService
...
-    private func getLastManagedTimeSlot() -> AnyObject?
+    private func getLastManagedElement() -> NSManagedObject?
     {
         let managedContext = self.getManagedObjectContext()
         
         let request = NSFetchRequest<NSFetchRequestResult>()
-        request.entity = NSEntityDescription.entity(forEntityName: timeSlotEntityName, in: managedContext)!
+        request.entity = NSEntityDescription.entity(forEntityName: self.getEntityName(), in: managedContext)!
         request.fetchLimit = 5
         request.sortDescriptors = [NSSortDescriptor(key: "startTime", ascending: false)]
```
amulware: This uses a specific key and order in a generic service.

teferi/Services/DefaultTimeSlotService.swift
willsb: Yup :)

willsb:
@amulware Please review and squash this is everything is okay :)

teferi/Services/DefaultTimeSlotService.swift
amulware: Why?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -2,188 +2,151 @@ import CoreData
...
         catch
         {
-            self.loggingService.log(withLogLevel: .error, message: "No TimeSlots found")
+            self.loggingService.log(withLogLevel: .error, message: "No \(self.entityName)s found")
             return nil
         }
     }
     
-    private func endPreviousTimeSlot(atDate date: Date) -> Bool
+    private lazy var entityName : String =
```
amulware: I'm still not convinced this should be in this service. If we ever change the class name of the model, this will break horribly, no? Should this not be in the adapter, similar to the names of the property keys?

teferi/Services/CoreDataPersistencyService.swift
```diff
@@ -2,188 +2,151 @@ import CoreData
...
         
         request.entity = entity
         request.predicate = predicate
         
         do
         {
-            guard let managedTimeSlot = try managedContext.fetch(request).first as AnyObject? else { return false }
-            managedTimeSlot.setValue(category.rawValue, forKey: "category")
+            guard let managedElement = try managedContext.fetch(request).first as AnyObject? else { return false }
+            updateFunction(managedElement)
```
amulware: This is still updating the managed object, exposing it to the outside, while every other interface only exposes T itself. Is there some reasoning for this?

teferi/Services/CoreDataPersistencyService.swift
willsb: If we change the entity name it will corrupt existing entries nevertheless, so I don't see why we should repeat this line of code when this does the trick

amulware: Do you mean that we cannot hardcode a custom string name, but HAVE TO use the class name?

Issue #136, opened 2016-11-22
Ask feedback option


pe0ny:
Add feedback option so users can contact us directly from the app.

![timeline - today 1x](https://cloud.githubusercontent.com/assets/5064445/20527841/9045349e-b0d2-11e6-9417-f1b5495c70aa.png)

amulware:
Do we have more details on what happens when the user hits that button?

In the simplest case, we can open an email-sending form, with attached log files, to the support email.
Great about that is that we then get the user's email and can answer them if we want to.

Alternatively we can have a custom dialog of some sort, which would require an endpoint from backend.

Or we could do a middle ground sort of thing: A custom form, and when the user hits 'send' it opens an email sending form with the feedback filled in (and they have to hit 'send' again).

If we have a custom form we could also include an emotion-picker like in the mobile app, asking the user if they feel :) :| or :( and asking them to rate the app if they like it.

pe0ny:
Yes, let's keep it simple and open an email form with attached log files.

PR #137, opened 2016-11-23
:sparkles: Feedback functionality

soundsmitten:
Feedback button in navigation bar and functionality using a generic form, per issue #136


teferi/ViewControllers/MainViewController.swift
```diff
@@ -249,4 +255,53 @@ class MainViewController : UIViewController
...
+        dismiss(animated: true)
+    }
+    
+    //Setup compose email form
+    func composeEmail()
+    {
+        //Check if email is configured on device
+        guard MFMailComposeViewController.canSendMail() else
+        {
+            let alert = UIAlertController(title: "Can't send email.", message: "Mail is not configured on his device. Either set it up in \"Settings -> Mail, Contacts, Calendars\" or email support@toggl.com using your preferred email client", preferredStyle: .alert)
```
willsb: Did you check this copy with Heidi? The "his device" part sounds kinda off :P

soundsmitten: No I haven't. I'll talk to her.

soundsmitten:
Wording for the prompts is now corrected.

willsb:
Nice work on this one 💃

amulware:
What is the status of this? Why was it approved if significant changes were expected? Is it now done?

teferi/Services/Protocols/FeedbackService.swift
```diff
@@ -0,0 +1,6 @@
+import UIKit
+
+protocol FeedbackService
+{
+    func composeEmail(recipients: [String], subject: String, body: String, logURL: URL?, parentViewController viewController: UIViewController)
```
amulware: The recipients, subject and body are specific to the email-feedback implementation of this service and should not be part of the protocol.

soundsmitten:
For the bug @willsb mentioned:

> Scroll to a day in the past
> Click the send feedback button
> Cancel it and delete the draft
> It will scroll back to current day, but the header will still display the older day. 
> I believe it shouldn't scroll back to the current date at all.

Located the issue here. PagerViewController's `viewWillAppear` is called when the email draft closes. In `viewWillAppear`, the date displayed in the TimeLineViewController is always set to the current `Date()`. Not sure about where to fix this without getting really messy/hackish  (likely with a boolean somewhere).

amulware:
It seems like a bug in the first place that the pager can switch days without switching title. That should be fixed. In addition, the original design said that the pager only switches back to 'today' if it was not visible for at least an hour (it was meant to have the app switch to today when the user had it in the background over night, and then started it the next day). I'm not sure that is necessarily the correct fix for this though. Ideally it should only consider switching day in the case where the app was opened from the background (maybe have the pager subscribe to the appstate's active state).

Just some thoughts. ¯\_(ツ)_/¯

amulware:
While it is a solution, I don't like that latest commit. It makes no sense to have that dependency. The solution should be more general. Otherwise we'll have to inject even more things in the future, if we add more features like this.

soundsmitten:
Last change is more generic.

teferi/ViewControllers/PagerViewController.swift
```diff
@@ -78,11 +76,13 @@ class PagerViewController : UIPageViewController, UIPageViewControllerDataSource
             .addDisposableTo(disposeBag!)
         
         //Check if the feedback UI has closed
-        if self.feedbackService.hasStartedFeedback == false {
-            self.initCurrentDateViewController()
-        } else {
-            self.feedbackService.hasStartedFeedback = false
+        guard !self.feedbackUIClosing else
```
amulware: I'd like this refactored into an if-statement.

teferi/Services/Protocols/MailFeedbackService.swift
```diff
@@ -5,14 +5,13 @@ import MessageUI
...
     //MARK: Fields
-    private let recipients: [String]
-    private let subject: String
-    private let body: String
-    
-    private var feedbackStarted: Bool = false
+    private let recipients : [String]
+    private let subject : String
+    private let body : String
+    private var completionHandler: () -> ()
```
amulware: There is no need for this property.

amulware:
Superseded by #153

Issue #128, opened 2016-11-10
Display an end point to show when the last time slot ended


pe0ny:
Display an end point to user to make it more understandable where and when last time slot ended. This applies only previous days (not current day).

pe0ny:
Fade last time slot out (only the line) and add end time.

![timeline - superyesterday 1x](https://cloud.githubusercontent.com/assets/5064445/20255336/72ffe5e4-aa44-11e6-8595-7eb62ec180ab.png)

amulware:
Am I correct that the fade is only used for the last entry of a day, if that entry is stopped? And not for the last entry, if it is 'running'?

pe0ny:
@amulware Yes. Only for the last entry of a day if that entry is stopped.
And only for previous days since today's last entry is always running and we'll show a dot there.

soundsmitten:
Once b7a89aba3 is merged, code can be reused to implement the line fade.

PR #139, opened 2016-11-23
:construction: Display an end point to show when the last time slot e…

soundsmitten:
Per issue #128 
In progress because I'm unsure this is the best design, because it changed some good timeline table cell view signatures.


teferi/Views/TimelineCell.swift
```diff
@@ -71,15 +71,16 @@ class TimelineCell : UITableViewCell
     }
     
     /// Updates the label that displays the description and starting time of the slot
-    private func layoutDescriptionLabel(withStartTime startTime: Date, category: Category)
+    private func layoutDescriptionLabel(withTimeSlot timeSlot: TimeSlot, category: Category, lastInPastDay: Bool = false)
```
willsb: I don't see the need for this parameter to be optional.

soundsmitten:
Also need to change the call in some other places 😮

amulware:
This should probably be rebased or redone (cherry picked) on a new branch.

PR #144, opened 2016-11-24
iOS10 Notifications

### LINKED ISSUE #117: Use iOS9/10 advanced notifications
The notifications support 4 buttons, we could use either the most common categories here, or just all non-commute categories.

In addition, we should consider if it is worth using the custom ui control we can use in iOS10, to show the current timeline (or just the current slot even), maybe with the last unknown entry showing a "?" inside it, which then updates as the user selects one of the options.

willsb:
Closes #117


NotificationContentExtension/TimeSlotMiniView.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.colorView.layer.cornerRadius = colorView.frame.width / 2
+    }
+    
+    func setUI(_ color : UIColor, category name : String?, date dateString: String)
+    {
+        self.colorView.backgroundColor = color
+        
+        let attributedString = NSMutableAttributedString()
+        
+        if let name = name
```
amulware: This works? Maybe check against nil instead?

Issue #130, opened 2016-11-10
Check the navigation bar spacing


pe0ny:
Please check the navigation bar spacing. Superday logo and calendar icon should be vertically aligned. Though at the moment it seems there's too much padding between logo and bottom part of the navigation bar.

Current state:
<img width="360" alt="screen shot 2016-11-10 at 12 29 41" src="https://cloud.githubusercontent.com/assets/5064445/20173311/68ca90ca-a741-11e6-91cc-de04fc484d31.png">

How it should be:
<img width="318" alt="screen shot 2016-11-10 at 12 30 16" src="https://cloud.githubusercontent.com/assets/5064445/20173324/7726e182-a741-11e6-85b5-40b23d601640.png">

soundsmitten:
![screenshot of opera 11-23-16 3-05-39 pm](https://cloud.githubusercontent.com/assets/852801/20578723/59217830-b18e-11e6-94f7-67696d11b93a.png) As you can see with the guidelines, technically the Calendar Icon is Y-aligned with the Superday logo. However, visually, it needs some sort of tweak. I feel that the "rings" of the calendar should be above the Superday logo, with the top of the Calendar paper in line with the top of the logo. Probably would need to change an inset to make this work.

pe0ny:
@soundsmitten I guess vertical alignment is not the issue then. But there's still some extra "padding" or something...? I tried to mark this with pink overlays:

<img width="379" alt="screen shot 2016-11-24 at 11 02 45" src="https://cloud.githubusercontent.com/assets/5064445/20592390/d13977f8-b235-11e6-90d5-baf0a3ac161a.png">

amulware:
As far as I can tell there is also a significant difference in horizontal alignment between the superday-icon and the category icons.

PR #145, opened 2016-11-24
:triangular_ruler: Less bottom padding in navigation bar

soundsmitten:
Per issue #130 

> As far as I can tell there is also a significant difference in horizontal alignment between the superday-icon and the category icons.

@amulware Should the superday-icon be pushed rightwards to center align with the category icons, then?


amulware:
@soundsmitten Yap :)

soundsmitten:
e3105fd yields this result:

screenshot of simulator 11-24-16 9-00-29 am

PR #149, opened 2016-11-25
Use Constraints instead of Frames

### LINKED ISSUE #142: Use auto-layout instead of hardcoded frames
As noted by @amulware, cases where the iPhone screen resizes (such as during calls) might mess up the entire layout if we don't use AutoLayout. We should review the app and remove hardcoded frames.

willsb:
Closes #142


amulware:
Are these really the only usages of frames? I'm surprised.

Minor thing: It seems more of a 📐 than a 💄 to me. ¯\_(ツ)_/¯

amulware:
Any more we are going to do on this?

PR #152, opened 2016-11-25
Change start date opacity to 50%

### LINKED ISSUE #132: Change start date opacity
Start time opacity should be 50% instead of 100%. It was also mentioned here #57

soundsmitten:
Closes #132

### DIFF
--- a/teferi/Extensions/StringExtensions.swift
+++ b/teferi/Extensions/StringExtensions.swift
@@ -41,4 +41,39 @@ extension String
         
         return attrStr
     }
+    
+    /**
+    Gets a string that contains bold, regular, and alpha-adjusted text
+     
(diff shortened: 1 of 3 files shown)


amulware:
The `getBoldStringWithNonBoldTextWithAlpha` method is extremely specific and only used in a single place. I think we should instead just use two labels, styled in the appropriate way.

Your thoughts, @willsb?

willsb:
There's no reason to use two labels instead of one. Specially since this is in a cell, so every extra `UIView` gets multiplied by `n` where n = number of TimeSlots in a day. If you think this is too specific, this could be a static method elsewhere instead of an extension, but I'm against using two labels when one does the trick.

amulware:
We'll do it by separating the text into two labels according to the logical separation.

PR #154, opened 2016-11-29
Refactor color constants into separate file

amulware:
Background: The previous implementation of using extension properties on UIColor hid the original UIColor properties, which might lead to confusion and accidental misuse. As part of this refactor, I also removed duplicate color constants from the Category enum, and refactored it optimizing for the most common use case.


teferi/Extensions/UIColorExtensions.swift
```diff
@@ -0,0 +1,38 @@
...
+    convenience init(hexString: String)
+    {
+        let hex = hexString.hasPrefix("#") ? hexString.substring(from: hexString.characters.index(hexString.startIndex, offsetBy: 1)) : hexString
+        var hexInt : UInt32 = 0
+        Scanner(string: hex).scanHexInt32(&hexInt)
+        
+        self.init(hex: Int(hexInt))
+    }
+    
+    var asHexString : String
```
willsb: I don't like the usage of `as` when lego naming a property. `hexString` sounds like a better alternative. Another option would be to turn this into a method (I prefer the property, though).

teferi/Styles/Color.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+
+class Color
+{
+    @nonobjc static let white = rgb(255, 255, 255)
```
willsb: Is there a need for the `@nonobjc` annotation here?

amulware: If it isn't, why was it in the UIColor extension file?

willsb: Because it extended `UIColor`, which is exposed to the ObjC runtime. `Color`, being a pure Swift class, doesn't need it

PR #157, opened 2016-11-29
:sparkles: Adding Calendar View

chriskryn:


teferi/ViewControllers/MainViewController.swift
```diff
@@ -39,6 +39,25 @@ class MainViewController : UIViewController, MFMailComposeViewControllerDelegate
     private var addButton : AddTimeSlotView!
     private var permissionView : PermissionView?
     private var launchAnim : LaunchAnimationView!
+    private var _calendarViewController: CalendarViewController?
+    fileprivate var calendarViewController: CalendarViewController?
```
willsb: I think a `lazy` property would fit better here

teferi/ViewControllers/MainViewController.swift
```diff
@@ -169,20 +188,114 @@ class MainViewController : UIViewController, MFMailComposeViewControllerDelegate
...
-            [ TimelineViewController(date: today,
-                                     metricsService: self.metricsService,
-                                     timeSlotService: self.timeSlotService,
-                                     editStateService: self.editStateService) ],
-            direction: .forward,
-            animated: true,
-            completion: nil)
+        if let calendarViewController = self.calendarViewController
+        {
+            if calendarViewController.view.alpha == 0
```
willsb: I don't like the alpha checking here. Ideally, we should either have an `isVisible` property on the `CalendarViewController` or simply have a `ToggleVisibility()` method that does all the work. I prefer the later as it'd remove responsability from the MainViewController, but I'd like to see @amulware's idea on this.

teferi/ViewControllers/MainViewController.swift
```diff
@@ -169,20 +188,114 @@ class MainViewController : UIViewController, MFMailComposeViewControllerDelegate
...
+                default:
+                    break
+                }
+            })
+            .addDisposableTo(disposeBag!)
         
-        self.onDateChanged(date: today)
+    }
+    
+    func hideCalendar(calendarViewController: CalendarViewController,
```
willsb: 1 - Why is this public?
2 - Why receive an instance of the VC when you can simply access the field?

teferi/ViewModels/CalendarViewModel.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  CalendarViewModel.swift
+//  teferi
+//
+//  Created by Krzysztof Kryniecki on 11/29/16.
+//  Copyright © 2016 Toggl. All rights reserved.
+//
+
```
willsb: Remove this

teferi/Views/CalendarCellView.swift
```diff
@@ -0,0 +1,125 @@
+//
+//  CalendarCellView.swift
+//  teferi
+//
+//  Created by Krzysztof Kryniecki on 11/28/16.
+//  Copyright © 2016 Toggl. All rights reserved.
+//
+
```
willsb: Remove this

teferi/Views/CalendarCellView.swift
```diff
@@ -0,0 +1,125 @@
...
+import UIKit
+import JTAppleCalendar
+
+class CalendarCellView: JTAppleDayCellView
+{
+
+    @IBOutlet weak var dateLabel: UILabel!
+    @IBOutlet weak var activityView: UIView!
+    let lightGreyColor = UIColor(colorLiteralRed: 206.0/255.0,
```
willsb: This should be included in the `Colors` class

amulware:
Tried checking out and building this, but it fails with:
```teferi/ViewControllers/CalendarViewController.swift:97:14: Value of type 'JTAppleCalendarView' has no member 'scrollToPreviousSegment'
teferi/ViewControllers/CalendarViewController.swift:102:14: Value of type 'JTAppleCalendarView' has no member 'scrollToNextSegment'```

chriskryn:
Have you done 'pod install' before building?
I added JTAppleCalendar to app pods.

chriskryn:
Ok I have found the problem, they must have released newer version of JTAppleCalendar and it doesn't have scrollToPreviousSegment and scrollToNextSegment. I have updated the podfile to use correct version.

amulware:
Work on this will continue on our own branch.

PR #158, opened 2016-11-30
Ignore locations when app is active

### LINKED ISSUE #127: Prevent app from generating time slots when app is active, or just went to background
How we get location updates when the app is active I do not know, but I suspect we get updates right away as the app goes into the background and we activate tracking, which causes the creation of new slots - often just after the user created a new slot manually.

slack_for_ios_upload_720-2

A general solution would be to always have tracking enabled, and simply ignore updates while the app is active. The second case might then become a non-issue. If it does not, we have to determine the criteria of how to filter for 'just went to background'.

Another option is to consider the user adding a new time slot as a sort-of location update, for purposes of timing and creation of new slots by the algorithm - this would mean that even if there are many updates (user is moving), if they set the category themselves recently, we will respect that until the phone comes to rest for long enough that a change in category is likely.

willsb:
Closes #127


teferi/AppDelegate.swift
```diff
@@ -70,7 +70,12 @@ class AppDelegate : UIResponder, UIApplicationDelegate
...
         self.locationService
             .locationObservable
-            .subscribe(onNext: self.trackingService.onNewLocation)
+            .subscribe(onNext: self.trackingService.onLocation)
+            .addDisposableTo(disposeBag)
+        
+        self.appStateService
+            .appStateObservable
+            .subscribe(onNext: self.trackingService.onAppState)
             .addDisposableTo(disposeBag)
```
amulware: These diffs are so bad...

Issue #160, opened 2016-12-01
Add TimeService


amulware:
```swift
protocol TimeService { var now : Date }
```

Mostly used for testing. Make sure this is used everywhere instead of `Date()` to get the current time. Might also include convenience members like `midnightToday`.

PR #163, opened 2016-12-01
Stop commute retroactively

### LINKED ISSUE #123: Stop commute retroactively on app start with no recent location updates
If the app starts (in the foreground OR from a notification, see #122), and the last location update is more than 25 minutes in the past, and the last timeslot is a commute slot, we should stop that slot at the last location update, and create a new one running since then. The user should not notice this change as it will happen before the UI is shown.

amulware:
Closes #123

Notes:
- This code does not take into account overnight behaviour of the app. This might not be needed and might need testing.
- In general, it seems to me that the tracking algorithm should be responsible for overnight behaviour in either case, which would make this a lot more predictable (and better unify concerns).
- The new behaviour has no tests since it needs #160.
- We should also add tests to over-night behaviour.


teferi/Services/DefaultTrackingService.swift
```diff
@@ -49,8 +51,7 @@ class DefaultTrackingService : TrackingService
         
         let currentTimeSlot = self.timeSlotService.getLast()
         
-        let difference = currentLocationTime.timeIntervalSince(previousLocationTime)
-        if (difference / 60) < 25.0
+        if isCommute(now: currentLocationTime, then: previousLocationTime)
```
willsb: I find your lack of `self.` disturbing

amulware:
I'm not happy with that last change. We should have a better way of ensuring that the data is valid on first startup. I propose a method we can call at the beginning of each method that will ensure this. The same method could also create the very first time entry ever.

PR #176, opened 2016-12-07
Fix tree

amulware:


amulware:
Would like to merge this to fix the git tree (i.e. get rid of the dangling tag). In the future we should probably make sure to always merge branches with tags properly.

PR #179, opened 2016-12-07
Model upgrade

### LINKED ISSUE #178: Add timeslot migration
Because of smart-guess addition

willsb:
Closes #178 

Since our changes qualify for a [Lightweight Migration](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CoreDataVersioning/Articles/vmLightweightMigration.html), all we need to do is ask Core Data to figure the migration and it just works®


teferi/AppDelegate.swift
```diff
@@ -223,7 +223,12 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         var failureReason = "There was an error creating or loading the application's saved data."
         do
         {
-            try coordinator.addPersistentStore(ofType: NSSQLiteStoreType, configurationName: nil, at: url, options: nil)
+            let options = [
+                NSMigratePersistentStoresAutomaticallyOption: true,
+                NSInferMappingModelAutomaticallyOption: true
+            ]
```
amulware: Will we run into problems with this in the future, should we add custom non-trivial migration logic?

willsb: Nope. If we ever add a complex model change, we have to go through the specific and custmo migration progress, but that won't be a problem

Issue #106, opened 2016-10-27
Write exhaustive unit tests for location algorithm


amulware:
Needs more. Can't be trusted.

Also need to find a way to test for the influence of other parts of the app on the algorithm if possible.

amulware:
Blocked by #160

amulware:
Once this is done, we should make sure to keep the tracking tests up to date at all times.

PR #181, opened 2016-12-08
🐛 Cancel notifications when guessing category for new slot

amulware:
Due to early return, notifications where not cancelled before.


teferi/Services/DefaultTrackingService.swift
```diff
@@ -52,13 +52,16 @@ class DefaultTrackingService : TrackingService
         
         let currentTimeSlot = self.timeSlotService.getLast()
         
+        let scheduleNotification : Bool
```
willsb: Why not initialize to true and then only change in the `else` branch?

amulware: I consider this much nicer since it forces each branch to consider whether or not it wants the notification to be scheduled (and nobody can change it after the fact, as would be the way with `var`). I consider this a very valid and good use of the `let` statement, signifying that the value does not change after being set.

amulware:
Should add tests for this.

amulware:
@willsb please approve or answer comment above.

I did not actually cover the case that this fixes with tests, since it relies on testing smart guesses. I added it to  #106 instead.

amulware:
Feel free to squash whenever.

PR #184, opened 2016-12-09
🌍 Ignore similar locations

### LINKED ISSUE #85: Complicate location algorithm
We should consider how we can use
- location
- speed
- ??

of the location updates to improve upon the existing location algorithm.

amulware:
Closes #85 

This includes appropriate tests, as well as a major refactor to the tracking tests since these now have to take into consideration location updates with proper coordinates.

Note that I hardcoded a threshold of 50 meters for the time being. As far as I can tell there is no objectively correct value here, and we need to look at actual data to determine if this or another value is best (Same with the several other constants we are using at the moment).

I suggest review per commit.


teferiTests/Services/TrackingServiceTests.swift
```diff
@@ -126,20 +115,21 @@ class TrackingServiceTests : XCTestCase
...
             .forEach(self.trackingService.onLocation)
         
         let allTimeSlots = self.timeSlotService.getTimeSlots(forDay: self.noon)
         let commutesDetected = allTimeSlots.filter { t in t.category == .commute }
         
-        expect(allTimeSlots.count).to(equal(5))
+        expect(allTimeSlots.count).to(equal(4))
         expect(commutesDetected.count).to(equal(2))
-        expect(allTimeSlots[3].startTime).to(equal(dates[4]))
+        expect(allTimeSlots[1].startTime).to(equal(dates[0]))
```
willsb: Any reason why this is not a loop?

teferiTests/Services/TrackingServiceTests.swift
```diff
@@ -166,29 +156,53 @@ class TrackingServiceTests : XCTestCase
...
+    
+    // Helper methods
+    
+    @discardableResult func setupFirstTimeSlotAndLastLocation(minutesBeforeNoon : Int) -> TimeSlot
+    {
+        return self.setupFirstTimeSlotAndLastLocation(
+            minutesBeforeNoon: minutesBeforeNoon, slotCategory: .unknown, wasSetByUser: false)
+    }
+    
+    @discardableResult func setupFirstTimeSlotAndLastLocation(
```
willsb: Giving default values for this function seem to work better, IMO

teferiTests/Services/TrackingServiceTests.swift
amulware: Would not really be any less code, not to mention that the two sequences of numbers are not trivially related.

teferiTests/Services/TrackingServiceTests.swift
amulware: I want the user to either specify both or neither of the optional arguments. Hence a second overload.

PR #188, opened 2016-12-12
🐛 Fix smart guess adapter crash on save

### LINKED ISSUE #187: Crash in smart guessing system
This occurs every time a time slot is edited or created, crashing the app, and causing time entry addition to not work at all.

```Fatal Exception: NSInvalidArgumentException
Unacceptable type of value for attribute: property = "category"; desired type = NSString; given type = _SwiftValue; value = teferi.Category.friends.
specialized SmartGuessModelAdapter.setManagedElementProperties(fromModel : SmartGuess, managedObject : NSManagedObject) -> ()
```

amulware:
Closes #187

This really should have been caught before. It appears the code in question was never actually tested.


amulware:
Squash whenever convenient.

PR #204, opened 2016-12-13
Use SwiftGen for type safety

willsb:
[Swift Gen](https://github.com/AliSoftware/SwiftGen) generates code for us. It basically analyzes the assets, localizable string and Storyboards and gives us type safety in lieu of using strings.


teferi/Services/PostiOSTenNotificationService.swift
```diff
@@ -91,19 +91,19 @@ class PostiOSTenNotificationService : NotificationService
     {
         let food = UNNotificationAction(
             identifier: Category.food.rawValue,
-            title: Category.food.rawValue.capitalized.translate())
```
amulware: Will the raw value change based on language? Or why does this not need to be translated?

teferi/AppDelegate.swift
```diff
@@ -113,8 +113,8 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         
         self.window = UIWindow(frame: UIScreen.main.bounds)
         
-        let storyboard = UIStoryboard(name: "Main", bundle: nil)
-        let mainViewController = storyboard.instantiateViewController(withIdentifier: "Main") as! MainViewController
+        let mainViewController = StoryboardScene.Main.instantiateMain()
```
amulware: I wish this would not have a method for each VC, but instead a property with an `instantiate` method, or similar.

teferi/Services/PostiOSTenNotificationService.swift
willsb: Currently we don't even have those values as Localizable strings, so this was not being translated in first place. When we start thinking about localizing superday we'll need to include those as I10n view

teferi/AppDelegate.swift
willsb: We could always write our own template, but I believe this is already good enough and far better than using raw strings

PR #205, opened 2016-12-13
📝 Add SuperFlow documentation

amulware:
Complete as far as I can tell, but please review. Again lengthy, but it seems necessary.


willsb:
Nice work on this one, Paulie :)

PR #210, opened 2016-12-15
Fix overlapping slots glitch

### LINKED ISSUE #126: Overlapping time slots
It appears this can still happen. The data seems to be correct, as when the app is relaunched it shows the correct times, but during the day itself, the timeline appears to not be updated correctly.

Screenshots:
slack_for_ios_upload_720

slack_for_ios_upload_720-1

willsb:
Closes #126


teferi/ViewModels/TimelineViewModel.swift
```diff
@@ -78,4 +96,11 @@ class TimelineViewModel
             self.timeSlotUpdatingVariable.value = index
         }
     }
+    
+    private func filterRefreshStates(_ appState: AppState) -> Bool { return appState == .active }
+    
+    private func mapIntoVoid(_ appState: AppState) -> Void
```
amulware: This method has a name that does not really describe what it is doing.

teferi/ViewModels/TimelineViewModel.swift
```diff
@@ -78,4 +96,11 @@ class TimelineViewModel
             self.timeSlotUpdatingVariable.value = index
         }
     }
+    
+    private func filterRefreshStates(_ appState: AppState) -> Bool { return appState == .active }
```
amulware: Why not just use a closure instead of this method?

teferi/ViewModels/TimelineViewModel.swift
```diff
@@ -17,7 +19,17 @@ class TimelineViewModel
     let isEditingObservable : Observable<Bool>
     let timeSlotUpdatingObservable : Observable<Int>
     let timeSlotCreationObservable : Observable<Int>
-    
+    private(set) lazy var refreshScreenObservable : Observable<Void> =
```
amulware: Why isn't this initialised in the constructor like the other observables?

willsb: One can't reference `self.` before `init`ing every constant field, but this one requires a reference to `self.`, hence needs to be lazy

PR #212, opened 2016-12-19
Rebranding to Superbeta

willsb:


willsb:
Going for `Supertoday` instead

PR #213, opened 2016-12-19
Rebrand to Supertoday

willsb:


teferi/AppDelegate.swift
```diff
@@ -36,7 +36,7 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         self.editStateService = DefaultEditStateService()
         self.loggingService = SwiftyBeaverLoggingService()
         self.locationService = DefaultLocationService(loggingService: self.loggingService)
-        self.feedbackService = MailFeedbackService(recipients: ["support@toggl.com"], subject: "Superday feedback", body: "")
```
amulware: This string should come from the translation files in the future.

PR #220, opened 2016-12-24
Fix duplication of pod scripts

pixyzehn:
I found duplication of scripts in build phases.

I guess each people may use a different version of CocoaPods, so I tried to use CocoaPods with a Gemfile.
Also, I regenerated scripts generated by CocoaPods like [CP] files.

See also: https://guides.cocoapods.org/using/a-gemfile.html


amulware:
You know this better than me, @willsb. Comment/review would be appreciated.

willsb:
Nice work on this one, @pixyzehn, thanks a lot :)

pixyzehn:
Thank you for your review :)

PR #630, opened 2017-06-16
HK remove short commutes TTL

### LINKED ISSUE #513: HK TTL: Remove short commutes
Let's set the bar at 5 minutes. Any commutes shorter than that we can remove.

I don't think that any additional processing of the TTL is needed afterwards, so this 'simple change' should be fine.

sofokli1:
Closes #513


teferi/Pipeline/HealthKitPump.swift
```diff
@@ -90,6 +98,27 @@ class HealthKitPump : Pump
...
     
+    private func removeSmallCommutesTimeSlots(from timeSlots: [TemporaryTimeSlot]) -> [TemporaryTimeSlot]
+    {
+        return timeSlots.enumerated().filter
+            { currentIndex, timeSlot in
+                guard timeSlot.category == .commute else { return true }
+                
+                let nextIndex = timeSlots.index(after: currentIndex)
+                
+                guard nextIndex < timeSlots.endIndex else { return true }
```
Odrakir: Not that it's wrong, but... Why are you using `endIndex` and `index(after:)` instead of plain old `count` and `+1`?

sofokli1: same logic as the above func. that func was a log time ago. maybe made sense for line 111?!

Odrakir: That line would work the same with the normal stuff... `index(after:)` and `endIndex`make sense if you are dealing with other kind of sequences, not necessarily arrays, but this is not the case and they are only making the code harder to read. I would change that. What do you think?

sofokli1: how long did it take you to understand the code? what a bout the function above that is the same and has passed review already? to me the code looks clear, the naming seems correct. i don't see a problem

PR #636, opened 2017-06-19
Smart guess

### LINKED ISSUE #595: Check if smart-guess algorithm is working correctly
It appears that at least for some (most? all?) people no categories are guessed (except commute, but this is a special case).

We have to investigate and figure out what is going on with this. If there is a bug, fix it. If there is not, let's see if we can improve it somehow (without unreasonable effort) to work better again.

Odrakir:
I removed the smartGuesses getting from the location pump and moved it to its own pipe. Also I'm adding some logging, see if we find what's going on.

This doesn't fix any issue with smart guess but it helps with logging and some separation of concerns.


teferi/Models/TemporaryTimeSlot.swift
```diff
@@ -20,6 +20,15 @@ extension TemporaryTimeSlot
         self.location = nil
     }
     
+    init(start: Date, category: Category)
```
sofokli1: wouldn't it make sense to set `category: Category = . unknown ` and remove the above init?
also if you add more default parameters instead of 4 inits you could have only 2 (1 instead of the first 3 and 1 for the last that calls the first 1 with the smart guess thing that is doing there.

if you do not like the suggested way, what about removing the body of the above init and just call this init from there like `init(start: start, category: .unknown)`

i think there is too much code duplication as is. if go with the suggestion or in an other way i leave it up to you.

MyPlayground.playground/Contents.swift
```diff
@@ -0,0 +1,49 @@
+//: Playground - noun: a place where people can play
```
sofokli1: remove the playground

teferi/Pipeline/SmartGuessPipe.swift
```diff
@@ -0,0 +1,26 @@
...
+    {
+        return timeline
+            .map(guessCategory)
+    }
+    
+    private func guessCategory(timeSlot: TemporaryTimeSlot) -> TemporaryTimeSlot
+    {
+        guard timeSlot.category == .unknown,
+            let location = timeSlot.location,
+            let smartGuess = smartGuessService.get(forLocation: location.toCLLocation()) else {
```
sofokli1: the `else` need to go a line below

Odrakir: Ugh! It's awful, but... ok.

Issue #642, opened 2017-06-21
Strange glitch in daily top bar


sofokli1:
may be related to #641 

![jun-21-2017 17-31-26](https://user-images.githubusercontent.com/885122/27392676-8d3c11fc-56a7-11e7-8cdc-4624ad13ab84.gif)

sofokli1:
i remember this to be fixed

Issue #641, opened 2017-06-21
Strange glitch in timeline


sofokli1:
As seen in the gif below, when scrolling to the top and then scrolling down a bit and not removing the finger from the screen some strange movements happen.
![jun-21-2017 17-27-27](https://user-images.githubusercontent.com/885122/27392537-380d022c-56a7-11e7-90e4-d429ebfbd618.gif)

sofokli1:
Here i just grab the work slot that starts at 5:36 PM and scroll up and down a few times, i did not relese my finger and in a few seconds the glitch happens
![jun-21-2017 17-39-34](https://user-images.githubusercontent.com/885122/27393143-be9c1ade-56a8-11e7-8667-31da272b2056.gif)

Odrakir:
Yeah, I've seen this before... it only happens in today's timeline, right?

sofokli1:
I think so, can't reproduce in other day

sofokli1:
I think this was fixed + can't reproduce

sofokli1:
i am seeing something similar again

amulware:
(Let's not delay the current release because of this. A fix would still be very welcome.)

sofokli1:
This shows how the issue appear after expanding a slot. The `scroll to top` behaviour in the end of the gif happens without touches.

![aug-09-2017 15-07-40](https://user-images.githubusercontent.com/885122/29120780-8df78b74-7d14-11e7-8679-4a9e328a3f40.gif)

Odrakir:
After all changes to the timeline, is this still a thing, @sofokli1? Can it be closed?

sofokli1:
I still see it. I think it is related to [this line here](https://github.com/toggl/superday/blob/develop/teferi/UI/Modules/Timeline/TimelineViewModel.swift#L56)

PR #644, opened 2017-06-22
Polish Daily Summary bar

### LINKED ISSUE #629: Polish daily timeline
The shadow only appears after the first category circle is pushed almost entirely under the bar already. I think it should appear as soon as the first cell starts disappearing.

Also, when switching to a different day that is scrolled to the top, the timeline instantly resets to maximum height. It would be awesome, if instead if would animate. (And same for the shadow, having it animate away, instead of disappearing instantly)

Odrakir:
Fixes #629 and #642


sofokli1:
i still see issue #641

Odrakir:
I'm sorry, I meant #642

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
```diff
@@ -17,32 +19,41 @@ class DailySummaryBarView : UIView
...
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        shadowView.layer.shadowPath = UIBezierPath(rect: shadowView.bounds).cgPath
+    }
+    
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        return backgroundView.hitTest(point, with:event)
     }
     
     func resize(by size: CGFloat)
```
sofokli1: the `size` parameter name is a bit misleading since it is not a `CGSize`

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
```diff
@@ -4,6 +4,8 @@ import UIKit
 class DailySummaryBarView : UIView
 {
     private let barView = BarView()
+    private let backgroundView = UIView()
+    private let shadowView = UIView()
```
sofokli1: why do you need this many views?

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
```diff
@@ -4,6 +4,8 @@ import UIKit
 class DailySummaryBarView : UIView
 {
     private let barView = BarView()
+    private let backgroundView = UIView()
```
sofokli1: this view seems to behave more like a container than a background. am i right?

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
```diff
@@ -81,6 +95,43 @@ class DailySummaryBarView : UIView
...
+    private func resetHeight()
+    {
+        barHeight = 10
+        height = 34
+
+        backgroundView.snp.updateConstraints { make in
+            make.height.equalTo(height)
+        }
+        
+        barView.snp.updateConstraints { make in
```
sofokli1: if `barView` has constraints based on `backgroundView`. should changing the `backgroundView` height be enough to change the `barView` height?

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
Odrakir: I need to be able to animate the position of the container and the opacity of the shadow independently. If I put the shadow in the container it messes up the animation and it jumps around.

teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
Odrakir: This was William's code, but: barView height is not equally proportional to the container height in both states of the header.

PR #645, opened 2017-06-23
Animate daily summary bar when changing dates

Odrakir:


teferi/UI/Modules/Timeline/Views/DailySummaryBarView.swift
```diff
@@ -186,4 +208,27 @@ class BarView: UIView
...
+    
+    private func animateTransition()
+    {
+        offset = Double(frame.width)
+        nextOffset = 0
+        self.displayLink = CADisplayLink(target: self, selector: #selector(BarView.animateOffset))
+        self.displayLink?.add(to: .current, forMode: .commonModes)
+    }
+    
+    var nextOffset: Double = 0
```
sofokli1: this var should be at the top of the class

amulware:
@sofokli1 No, we can merge these changes. We can always revert, if there are objections.

PR #650, opened 2017-06-28
:bug: Fix tests

### LINKED ISSUE #585: Check MainViewModel test
Check if `testSmartGuessIsStrikedIfCategoryWasWrongOnUpdateTimeSlotMethod`is doing the right thing, just cause I don't think it is... The SmartGuess comes from the previous TS (.leisure)

sofokli1:
Closes #585

The mock defaultSmartGuessService was missing the logic for increasing the smartGuess id


teferiTests/Mocks/MockSmartGuessService.swift
```diff
@@ -23,6 +23,8 @@ class MockSmartGuessService : SmartGuessService
         let smartGuess = SmartGuess(withId: id, category: category, location: location, lastUsed: Date())
         smartGuesses.append(smartGuess)
         
+        incrementSmartGuessId()
```
Odrakir: Not sure if you need that indirection. Also, while you are on it, can you rename `id` to something more descriptive like `smartGuessId` or whatever?

teferiTests/ViewModels/MainViewModelTests.swift
```diff
@@ -142,6 +145,8 @@ class MainViewModelTests : XCTestCase
                                                         smartGuess: SmartGuess(withId: 0, category: .food, location: location, lastUsed: Date()),
                                                         location: location)!
         
+        expect(timeSlot.categoryWasSetByUser).to(beFalse())
```
Odrakir: Is this related?

sofokli1: the check in line 152 is useless unless we know for sure that it was not set by user in the first place

teferiTests/Mocks/MockSmartGuessService.swift
sofokli1: the func name and logic comes from the original class that is trying to mock. i will rename the `id` but don't see anything else that can be misunderstood for, apart from the obvious.

Odrakir: `incrementSmartGuessId` is a private function that doesn't need to follow the naming of the class we are trying to mock. And what I'm saying is that the indirection is unnecessary. We have a function called `incrementSmartGuessId` that only has one line that does exactly that. So you might as well replace the function called by that line.

Any way, it's not a big deal, you can leave it like that if you want to.

teferiTests/ViewModels/MainViewModelTests.swift
Odrakir: Yes, but it's unrealted to the issue this PR fixes, right? Anyway, again, not a big deal, just something we should try not to do generally.

PR #651, opened 2017-06-28
:construction: Collapse all days into one when guessing

sofokli1:


Odrakir:
Eugene, please, add description or link to something in the PR.

How about, instead of completely discarding the day of the week, we give different weights for those two things?. Example: If the hour matches it weights whatever for the smart guess match, but if the day of the week also matches then it weights more.

sofokli1:
@Odrakir what do you think about this simple approach where i give importance if they are from the same group (weekdays, weekends)? if none is available from the same group the slot from the other group will still matter as they do now.

teferi/Extensions/DateExtensions.swift
```diff
@@ -108,9 +110,8 @@ extension Date
         return calendar.date(from: currentComponents)!
     }
     
-    func timeIntervalBasedOnWeekDaySince(_ date: Date) -> TimeInterval
+    func timeIntervalBasedOnSameDaySince(_ date: Date) -> TimeInterval
     {
-        let dayDifference = dayOfWeek - date.dayOfWeek
-        return timeIntervalSince(date.convert(calendarUnits: [ .year, .month, .day], sameAs: self).add(days: -dayDifference))
+        return timeIntervalSince(date.convert(calendarUnits: [ .year, .month, .day], sameAs: self))
```
Odrakir: But anyway, now that I think about it, neither of those seem to work. If I put this in a playground:

```swift
let calendar = Calendar.current
var components = DateComponents()
components.day = 15
components.month = 10
components.year = 2016
components.hour = 23
components.minute = 50
let date1 = calendar.date(from: components)!
let date2 = date1.addingTimeInterval(60*60)
let t = abs(date2.timeIntervalBasedOnSameDaySince(date1)) / 3600
```

t is 23, which means 23 hours, when it should be just 1 hour (60*60). So something is wrong when changing days...

teferi/Services/Implementations/Persistency/DefaultSmartGuessService.swift
```diff
@@ -177,9 +177,12 @@ class DefaultSmartGuessService : SmartGuessService
         let locationDifference = instance1.location.distance(from: instance2.location) / distanceThreshold
         accumulator += pow(locationDifference, 2)
 
-        let timeDifference = instance1.timeStamp.timeIntervalBasedOnWeekDaySince(instance2.timeStamp) / timeThreshold
+        let timeDifference = instance1.timeStamp.timeIntervalBasedOnSameDaySince(instance2.timeStamp) / timeThreshold
         accumulator += pow(timeDifference, 2)
         
+        let weekendDifference = instance1.timeStamp.isWeekend == instance2.timeStamp.isWeekend ? 1.0 : 0.0
```
Odrakir: Maybe I don't understand the logic behind this, but... you are just checking if they both are days belonging to a weekend or not, so not really checking if they are the same day or how many days apart they are, right?

Please, explain if I'm missing something.

In my mind we would have two more variables added to the accumulator, or one of the two, depending on how the algorithm works...
- difference in hours (based on the same day)
- difference in week days

sofokli1: if both are weekend they get 1.0 and if both are weekdays they get 1.0. if one is weekend and one is weekday it gets 0.0

regarding:
difference in hours (based on the same day)
difference in week days
- how would you convert that number from each of the above to a score from 1.0 to 0.0 ?
- should a linear change from 1.0 to 0.0 be equal to a linear change of the above or should it be exponential or something else?

we can add these and we should add as many as possible but without data is hard to test and know if it is helping or not.

teferi/Extensions/DateExtensions.swift
sofokli1: i see your point. there is an issue with the implementation in this PR and also with the live implementation. Working on fix for the live and will apply the same logic later to this one as well.

teferi/Services/Implementations/Persistency/DefaultSmartGuessService.swift
Odrakir: Difference in hours goes from 0 to 12
Difference in days goes from 0 to 4

Both are then easy to normalize.

amulware:
Is there an issue for this? Is it a bug fix or a change of behaviour? If the latter, what is the reasoning?

PR #652, opened 2017-06-28
:sparkles: Remove commute smartguess

### LINKED ISSUE #550: Consider not saving commute smart guesses
Mostly, right now if there are commute smart guesses in locations the user goes to a lot, the whole timeline might turn blue for large parts of the day, which looks very bad.

In general, we may want to change the way we think about movement entirely, and instead of treating it as a category like everything else, treat it as a special thing (at least UI wise).

We already talked about displaying 'commutes' differently, so this would play well with that.

sofokli1:
Closes #550


amulware:
I'm not 100% sure we'll want this behaviour in the long term, but at least for now I think it's an improvement.

PR #654, opened 2017-07-03
Move onboarding logic to view models

### LINKED ISSUE #580: Move onboarding logic to view models
The idea is to not have direct access to the services from the view controllers, everything should go trough the view models, which will have inputs and outputs. This will make code easier to test.

Odrakir:
Fixes #580 

I created a VM for all the pages, we still could go a little bit further, but I think it's enough for now.


teferi/UI/Modules/On Boarding/OnboardingPageViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+        guard !settingsService.hasLocationPermission else { return }
+        
+        locationManager = CLLocationManager()
+        locationManager.delegate = self
+        locationManager.requestAlwaysAuthorization()
+    }
+    
+    private func time(from hourString: String) -> Date
+    {
+        guard let hours = Int(hourString.components(separatedBy: ":")[0]),
```
sofokli1: how about caching `hourString.components(separatedBy: ":")`

teferi/UI/Modules/On Boarding/OnboardingPageViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+    func requestLocationAuthorization()
+    {
+        guard !settingsService.hasLocationPermission else { return }
+        
+        locationManager = CLLocationManager()
+        locationManager.delegate = self
+        locationManager.requestAlwaysAuthorization()
+    }
+    
+    private func time(from hourString: String) -> Date
```
sofokli1: since you are getting both hours and minutes from the same string maybe `timeString` would be a better name? hours seems a bit misleading.

teferi/UI/Modules/On Boarding/OnboardingPage3.swift
```diff
@@ -14,43 +13,30 @@ class OnboardingPage3 : OnboardingPage, CLLocationManagerDelegate
     
     override func startAnimations()
     {
-        disposeBag = disposeBag ?? DisposeBag()
+        disposeBag = DisposeBag()
```
Odrakir: I'm try to simplify what was already there. The conditional check used to create a `disposeBag` if it was nil, and the `disposeBag` used to get set to nil in `finish`. I think we can remove the conditional check and setting the `disposeBag` to nil in `finish` if we just always re-create it here. The fact that it's also created two more times is the weird thing. I'll remove this line and everything should work the same and make more sense, the `disposeBag` can't be nil and you just replace it when you want things to get disposed.

teferi/UI/Modules/On Boarding/OnboardingViewModel.swift
```diff
@@ -20,4 +20,14 @@ class OnboardingViewModel
         self.appLifecycleService = appLifecycleService
         self.notificationService = notificationService
     }
+    
+    func viewModel(forPage: Int) -> OnboardingPageViewModel
```
Odrakir: Yep, I started sending that param thinking I was going to need it, but didn't. I'll remove it.

teferi/UI/Modules/On Boarding/OnboardingPageViewModel.swift
Odrakir: 👍

teferi/UI/Modules/On Boarding/OnboardingPageViewModel.swift
Odrakir: I don't think it's worth the extra line outside of the guard. Makes the code messier and this won't impact performance.

PR #657, opened 2017-07-04
:bug: Fix date extension

sofokli1:
Fix timeInterval measurement when collapsing everything into a week

Feel free to try dates and break the logic. I tested with crazy dates for some hours and seems to work fine.


teferi/Extensions/DateExtensions.swift
```diff
@@ -108,9 +108,21 @@ extension Date
         return calendar.date(from: currentComponents)!
     }
     
-    func timeIntervalBasedOnWeekDaySince(_ date: Date) -> TimeInterval
+    func shortestTimeIntervalByCollapsingWeeksTogether(_ date: Date) -> TimeInterval
```
Odrakir: I would call this method something like `timeIntervalInWeek` and then I would add a comment explaining what it actually does, as it's one of those rare cases in which it's not clear by the name.

teferi/Services/Implementations/Persistency/DefaultSmartGuessService.swift
```diff
@@ -174,7 +174,7 @@ class DefaultSmartGuessService : SmartGuessService
         let locationDifference = instance1.location.distance(from: instance2.location) / distanceThreshold
         accumulator += pow(locationDifference, 2)
 
-        let timeDifference = instance1.timeStamp.timeIntervalBasedOnWeekDaySince(instance2.timeStamp) / timeThreshold
+        let timeDifference = instance1.timeStamp.shortestTimeIntervalByCollapsingWeeksTogether(instance2.timeStamp) / timeThreshold
```
Odrakir: Still not sure if this is going to work. You are only checking the difference in hours in the same week. Meaning if something happens on a Monday, the smartGuess on Tuesday, will have more weight than one on a Thursday, which it's not what we want.

sofokli1: this is only a fix for something that was broken and not a change in functionality

Odrakir: Oh, I didn't realize it was broken. Ok.

By the way, can you add some of those tests you did as unit tests?

teferiTests/Extensions/DateExtensionsTests.swift
```diff
@@ -74,4 +74,88 @@ class DateExtensionsTests : XCTestCase
         
         expect(date.ignoreTimeComponents()).to(equal(date2.ignoreTimeComponents()))
     }
+    
+    func testTimeIntervalInWeekMeasuresCorrectlyForSameDayInDifferentWeeks()
```
amulware: I'd prefer the test method names to specify what behaviour is expected.

teferiTests/Extensions/DateExtensionsTests.swift
```diff
@@ -74,4 +74,88 @@ class DateExtensionsTests : XCTestCase
...
+        components.year = 2017
+        components.hour = 00
+        components.minute = 50
+        let date1 = calendar.date(from: components)!
+        let date2 = date1.addingTimeInterval(-14*24*60*60 - 60*60)
+        
+        expect(date1.timeIntervalInWeek(date2)).to(equal(60*60))
+    }
+    
+    func testTimeIntervalInWeekReturnsThreeDaysOrLessDifferenceForDatesThatAreMoreThenThreeDaysApart()
```
amulware: Is the actual maximum of the function not 3.5 days? i.e. exactly half a week?

teferi/Extensions/DateExtensions.swift
```diff
@@ -108,9 +108,25 @@ extension Date
...
+    /// - Returns: The absolut value of the time difference in seconds
+    func timeIntervalInWeek(_ date: Date) -> TimeInterval
     {
-        let dayDifference = dayOfWeek - date.dayOfWeek
-        return timeIntervalSince(date.convert(calendarUnits: [ .year, .month, .day], sameAs: self).add(days: -dayDifference))
+        var absDayDifference = abs(dayOfWeek - date.dayOfWeek)
+        
+        if absDayDifference > 3
+        {
+            absDayDifference = absDayDifference - ( 2 * ( absDayDifference - 3 ) - 1 )
```
Odrakir: Couldn't you replace this by:
```swift
absDayDifference = 7 - absDayDifference
```
??

sofokli1:
@amulware is using midnight instead of noon and it is one of the main edge cases. at least 3 of these new tests would fail with the old implementation.

teferiTests/Extensions/DateExtensionsTests.swift
sofokli1: day wise you can only move in integer values, the rest is hours and minutes. From today (regardless of the hour) you can only go 3 days in the future and if you go one more you find yourself 3 days in the past. Here i am testing that the second date collapse in the same week (in a wrap around ways) correctly. The 0.5 day is just `.timeIntervalSince...`

sofokli1:
i am having trouble finding a logical reason why this test fails in bitrise but not locally. the only thing i can think of is to remove the test. Any suggestions?

PR #658, opened 2017-07-05
Testing ci

sofokli1:


sofokli1:
i am having trouble finding a logical reason why this test fails in bitrise but not locally. the only thing i can think of is to remove the test.

PR #659, opened 2017-07-05
Add a pipe to merge short slots

### LINKED ISSUE #594: TTL: Remove/merge small timeslots after merge
We should look at real data from at least several of us, to see what lengths are problematic.

After deciding the limit, we should add a pipe that merges small slots into larger adjacent ones after the merging pipe.

A key problem to solve is how to keep the meta-data of the changed slots in tact. We do not want to lose locations/categories/smartguesses and have to find a way to resolve conflicts.

Odrakir:
Fixes #594 

I inserted it after the one that merges mini commute slots, let's see how it works...


teferi/Pipeline/MergeShortSlotsPipe.swift
```diff
@@ -0,0 +1,138 @@
...
+        var shortestSlot = auxTimeline[shortestSlotIndex]
+        
+        while shortestSlot.duration! < shortTimeSlotThreshold
+        {
+            defer {
+                shortestSlotIndex = findShortestSlotIndex(auxTimeline)
+                shortestSlot = auxTimeline[shortestSlotIndex]
+            }
+            
+            guard shortestSlotIndex + 1 < auxTimeline.count,
```
Odrakir: The last slot won't change once we show it to the user, will it? And the pipeline runs when the app becomes active. Which means that last timeslot can't grow.

I'm growing the previous TS to match the deleted one end time, if that helps...

I have an idea of a way of doing this in which the current (last) day's timeline is always temporary until the day changes... But that's for another discussion.

amulware: The last slow will change, in that it's end time is not yet set, so it will likely grow above the 5m limit. What we lose is the split between two the slots (second to last, and last, i.e. current running), which I think is important.

Odrakir: Oh... you are right. But then, when adding a TS after that one, then we should decide if we should delete that TS or not (depending on its time).

I'll prevent it from being deleted and we can see about that other thing in another issue?

amulware: I don't think we can do much at this point, since the user has now seen it. But that's a minor thing I think. So yes, prevent it from being deleted here, and we can see if we can find a good solution down the road.

teferi/Pipeline/MergeShortSlotsPipe.swift
```diff
@@ -0,0 +1,138 @@
...
+                areConsecutive(shortestSlot, auxTimeline[shortestSlotIndex + 1])
+            else {
+                auxTimeline.remove(at: shortestSlotIndex)
+                auxTimeline[shortestSlotIndex - 1] = auxTimeline[shortestSlotIndex - 1].with(end: shortestSlot.end)
```
Odrakir: I was going to say that can happen, because you always check the timeline has more than 1 element and so, if that index is 0 it'll go into the `if`. But yes, it turns out it can happen if all timeslots in the timeline are too short. I'm fixing it adding a check for `auxTimeline.count > 1` in the `while`.

teferi/Pipeline/MergeShortSlotsPipe.swift
```diff
@@ -0,0 +1,138 @@
...
+        return auxTimeline
+    }
+    
+    private func findShortestSlotIndex(_ timeline: [TemporaryTimeSlot]) -> Int
+    {
+        var shortestDuration: TimeInterval = TimeInterval.greatestFiniteMagnitude
+        var shortestIndex: Int = 0
+        for (index, ts) in timeline.enumerated()
+        {
+            guard let duration = ts.duration else {
```
Odrakir: I'm guessing for the last slot, which doesn't have an end time. So we were actually already skipping that last slot, even if the tests didn't catch it (as in the tests all TS have end times). Anyway, I think the extra check doesn't hurt.

teferi/Pipeline/MergeShortSlotsPipe.swift
amulware: What 'if' do you mean? I'm pretty sure this case can happen if you have a timeline starting with a small slot followed by a gap (and then any number of slots) as well.

teferi/Pipeline/MergeShortSlotsPipe.swift
amulware: Ah, I see. I didn't realize that TTL.duration is nil if the end time is nil

teferi/Pipeline/MergeShortSlotsPipe.swift
Odrakir: You are right, I wasn't taking gaps into account.

teferi/Pipeline/MergeShortSlotsPipe.swift
```diff
@@ -0,0 +1,140 @@
...
+    {
+        guard timeline.count > 1 else { return timeline }
+        
+        var auxTimeline = timeline
+        var shortestSlotIndex = findShortestSlotIndexSkipingLast(auxTimeline)
+        var shortestSlot = auxTimeline[shortestSlotIndex]
+        
+        while shortestSlot.duration! < shortTimeSlotThreshold && auxTimeline.count > 1
+        {
+            defer {
```
Odrakir: I feel really awful puting that bracket in a new line after just one word... I'm sorry, I tried but I can't get used to this.

teferi/Pipeline/MergeShortSlotsPipe.swift
```diff
@@ -0,0 +1,140 @@
...
+                shortestSlotIndex = findShortestSlotIndexSkipingLast(auxTimeline)
+                shortestSlot = auxTimeline[shortestSlotIndex]
+            }
+            
+            guard shortestSlotIndex + 1 < auxTimeline.count,
+                areConsecutive(shortestSlot, auxTimeline[shortestSlotIndex + 1])
+            else {
+                auxTimeline.remove(at: shortestSlotIndex)
+                if shortestSlotIndex > 0 && areConsecutive(auxTimeline[shortestSlotIndex - 1], shortestSlot) {
+                    auxTimeline[shortestSlotIndex - 1] = auxTimeline[shortestSlotIndex - 1].with(end: shortestSlot.end)
```
Odrakir: Why you say so?

amulware: The one two lines above?

sofokli1: didn't see between the new line comments 🙈

sofokli1: why outside the if?

sofokli1: i see now, it makes sense

PR #661, opened 2017-07-06
:sparkles: Collapse time in one day

sofokli1:
I tried to keep the logic as simple as possible


teferi/Extensions/DateExtensions.swift
```diff
@@ -108,9 +108,12 @@ extension Date
         return calendar.date(from: currentComponents)!
     }
     
-    func timeIntervalBasedOnWeekDaySince(_ date: Date) -> TimeInterval
+    func timeIntervalForSameDaySince(_ date: Date) -> TimeInterval
```
Odrakir: How about `timeIntervalIgnoringDate`. I don't want to be picky, just trying to find the best name so we remember what this does in the future.

teferi/Services/Implementations/Persistency/DefaultSmartGuessService.swift
```diff
@@ -166,7 +166,7 @@ class DefaultSmartGuessService : SmartGuessService
             let smartGuessTimestamp = smartGuess.location.timestamp
             let locationTimestamp = location.timestamp
             
-            return abs(smartGuessTimestamp.timeIntervalBasedOnWeekDaySince(locationTimestamp)) <= self.timeThreshold
+            return abs(smartGuessTimestamp.timeIntervalForSameDaySince(locationTimestamp)) <= self.timeThreshold
```
Odrakir: The name of this method then, shouldn't be `isWithinTimeThresholdInNearByWeekDay` any more. How about, following my previous comment: `isWithinTimeThresholdIgnoringDate`?

teferi/Extensions/DateExtensions.swift
amulware: Is it worth including something like 'absolute' in the name, to make it clear that this does not return negative values?

teferi/Extensions/DateExtensions.swift
```diff
@@ -108,9 +108,12 @@ extension Date
...
     {
-        let dayDifference = dayOfWeek - date.dayOfWeek
-        return timeIntervalSince(date.convert(calendarUnits: [ .year, .month, .day], sameAs: self).add(days: -dayDifference))
+        let differenceIfPreviousDay = abs(self.ignoreDateComponents().timeIntervalSince(date.ignoreDateComponents().add(days: -1)))
+        let differenceIfSameDay = abs(self.ignoreDateComponents().timeIntervalSince(date.ignoreDateComponents()))
+        let differenceIfNextDay = abs(self.ignoreDateComponents().timeIntervalSince(date.ignoreDateComponents().add(days: 1)))
```
amulware: Might be nicer to cache the two times-without-date for readability, and performance?

PR #663, opened 2017-07-11
Change TimeSlot model to a struct

Odrakir:
This merges to a temporary branch so it eventually will add the collapsable timeline.

Had to remove some tests that made no sense with immutable models. Those tests may have to be put somewhere else (like persistency)


teferi/Models/TimeSlot.swift
```diff
@@ -15,20 +14,66 @@ class TimeSlot
     var categoryWasSetByUser : Bool
```
amulware: Shouldn't these all be `let`s?

teferi/Services/Implementations/Persistency/DefaultTimeSlotService.swift
```diff
@@ -95,22 +95,16 @@ class DefaultTimeSlotService : TimeSlotService
...
     {
-        guard canChangeCategory(of: timeSlot, to: category, setByUser: setByUser) else { return }
-        
         let predicate = Predicate(parameter: "startTime", equals: timeSlot.startTime as AnyObject)
         let editFunction = { (timeSlot: TimeSlot) -> (TimeSlot) in
-            
-            timeSlot.categoryWasSetByUser = setByUser
-            timeSlot.category = category
-            return timeSlot
+            return timeSlot.with(category: category, setByUser: true)
```
amulware: Can we assume true here? I'm afraid that we forget to re-add this argument should it be needed in the future.

teferi/Services/Implementations/Persistency/DefaultTimeSlotService.swift
```diff
@@ -95,22 +95,16 @@ class DefaultTimeSlotService : TimeSlotService
         return timeSlots
     }
     
-    func update(timeSlot: TimeSlot, withCategory category: Category, setByUser: Bool)
+    func update(timeSlot: TimeSlot, withCategory category: Category)
     {
-        guard canChangeCategory(of: timeSlot, to: category, setByUser: setByUser) else { return }
```
amulware: Why remove this check?

teferiTests/Mocks/MockSmartGuessPersistencyService.swift
```diff
@@ -22,9 +22,9 @@ class MockSmartGuessPersistencyService : BasePersistencyService<SmartGuess>
         return true
     }
     
-    @discardableResult override func update(withPredicate predicate: Predicate, updateFunction: @escaping (T) -> T) -> Bool
+    @discardableResult override func update(withPredicate predicate: Predicate, updateFunction: @escaping (T) -> T) -> T?
     {
-        return true
+        return nil
```
amulware: How come this doesn't blow the tests?

teferi/Models/TimeSlot.swift
Odrakir: Yes, let me refactor that struct a little bit. Also custom initializers should be in an extension so we still have the default one.

teferi/Services/Implementations/Persistency/DefaultTimeSlotService.swift
Odrakir: Since we changed how the timeline generation works the only way to change a TimeSlot's category is through user interaction, and it's always allowed.

amulware: The method also checked if the new category is different than the old one. Is that check not important?

teferi/Services/Implementations/Persistency/DefaultTimeSlotService.swift
Odrakir: As said above, right now it's the only way to update categories on TimeSlots, so `setByUser` is always true. I don't think we should keep old code just in case we need it in the future...

teferiTests/Mocks/MockSmartGuessPersistencyService.swift
Odrakir: Because it's never tested :D

Odrakir:
About the deleted tests... I suggest creating a new issue about this, because I'm finding more and more tests that either are not in the correct test class or span different classes.

The two I deleted here are `testTheUpdateTimeSlotMethodChangesATimeSlotsCategory` and `testTheUpdateMethodMarksTimeSlotAsSetByUser` which are part of the `MainViewModel` tests.

In my opinion those two tests don't belong here, as the view model should just take the view controller's command, do whatever checks it needs to do and forward it to the appropriate "service" (in this case it's a service or multiple of them, as I always said, it should be a single "Interactor"). So in any case it's that service, interactor or maybe even the persistency what should be tested for those tests.

These two cases showed up here because they stopped working with the changes towards immutability, but there are many more, that's why I think it's worth it to spend some time reviewing tests.

amulware:
I have no argument with regards to the interactors, or the tests being moved somewhere else. Please create appropriate issues, or take note of them in some other way so that we won't forget.

Otherwise, there is still one question above:

> The method also checked if the new category is different than the old one. Is that check not important?

Odrakir:
The category will always be different, as the EditView only has all other categories. Also, it wouldn't actually crash or anything if we "change" a TimeSlot to the same category. Nevertheless, I'm adding a guard just to be safe.

PR #669, opened 2017-07-12
Expands and collapses timeline items

### LINKED ISSUE #620: Collapse merged time slots
Visually collapse merged time slots with same category by default like this:
yesterday collapsed slots 1x

When user taps on the three dots, show visually merged time slots like we did before, but with option to collapse slots again:
yesterday merged time slots 1x

note! collapsed slot height: 20pt first slot, 3 dots 4pt height, 20pt last slot

Odrakir:
Fixes #620 minus the animation


Odrakir:
I wasn't sure about the height, and actually I'm still not sure. I think it should probably be some value between a max and a min, but yes, the min should be bigger than that screenshot, so the preceding and following lines show. Anyway, I'll fix it to the height of the designs and we can discuss it later.

I was leaving the editing for another PR as this is more view related.

Also the logic in expanding/contracting, but yeah, that should be fixed here, will do.

Are you reviewing code as well?

amulware:
I was only glancing over it. If you'd like my to read it properly, let me know and I'll do it tomorrow!

amulware:
The slot still collapses under almost all circumstances when editing.

Specifically.. it seems to collapse always when a slot in the group is edited out of it, and only stays open when a slot is 'added' to the group.

Can we still change that here? Or would you rather merge this first? (If the latter, feel free.)

Odrakir:
The logic behind it is we always follow the edited slot. So if you edit it out of a group and into the next one, for example. it'll close the current one and open the next one. If you edit it out of a group to be by itself, it'll close that group and nothing else. This behavior works for me, are you sure you want to keep the old group open?

amulware:
Ah, I see now what you mean. Yes, it works that way for me as well. Tough call... I feel that I'd like the group to stay open when removing a slot from it (what if the user tries to edit multiple out of the group?) But I see your point that it makes sense to follow the edited slot. So the group can really only stay open if the edited slot is now standing alone. And what if it was the 3rd slot in a group of 5...

Let's leave it like this, and see what @pe0ny thinks after testing it, and how it feels with animations.

PR #679, opened 2017-07-27
:sparkles: Welcome message (aka missing onboarding step)

### LINKED ISSUE #539: App doesn't show the last onboarding step
When installing the app and going through the onboarding view, the app should show this screen:
(atm, it shows it with out the 'That's it. Check back later..' part)

step5 1x

sofokli1:
Closes #539


amulware:
Where did this go, or was it never there?

sofokli1:
i am not sure my self

teferi/UI/Modules/Main/MainViewController.swift
```diff
@@ -108,6 +108,10 @@ class MainViewController : UIViewController, MFMailComposeViewControllerDelegate
         viewModel.showPermissionControllerObservable
             .subscribe(onNext: presenter.showPermissionController)
             .addDisposableTo(disposeBag)
+        
+        viewModel.welcomeMessageHiddenObservable
```
Odrakir: I think this is simpler and communicates intent better this way: 
```swift
viewModel.welcomeMessageHiddenObservable
            .bindTo(welcomeMessageView.rx.isHidden)
            .addDisposableTo(disposeBag)
```

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -30,6 +30,21 @@ class MainViewModel : RxViewModel
             .filterNil()
     }
     
+    var welcomeMessageHiddenObservable : Observable<Bool>
+    {
+        return Observable.of(
+            self.appLifecycleService.movedToForegroundObservable,
+            self.didBecomeActive,
+            self.beganEditingObservable.flatMap { _, _ in Observable<Void>.just() },
```
Odrakir: Instead of the flatMap, `self.beganEditingObservable.mapTo(())` is enough.

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -30,6 +30,21 @@ class MainViewModel : RxViewModel
...
+            self.timeSlotService.timeSlotCreatedObservable.flatMap { _ in Observable<Void>.just() } )
```
Odrakir: Instead of the flatMap, `self.beganEditingObservable.mapTo(())` is enough.

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -30,6 +30,21 @@ class MainViewModel : RxViewModel
             .filterNil()
     }
     
+    var welcomeMessageHiddenObservable : Observable<Bool>
+    {
+        return Observable.of(
+            self.appLifecycleService.movedToForegroundObservable,
+            self.didBecomeActive,
```
Odrakir: [Sorry this comment didn't get sent with the last review! (not sure what happened there, but it was still here in my browser)]

This observable sends a value when the viewModel activates (not the app). So this whole thing will get executed as soon as the message shows. Making the rest of the observables you are merging, useless for updating the `welcommeMessageHidden` setting.

Not sure if that was your intention, because it does work, but it seems a little confusing.

sofokli1: The observer should fire as it is doing. The merge the first time will return `false` any other time will return `true`. it should return `false` only the first time. If the VC is recreated for any other reason (together with its model) it will not show the welcome view.

PR #685, opened 2017-08-01
Fix calendar presentation being too slow

### LINKED ISSUE #591: Showing the calendar is a little slow
We should check this with the time profiler and see what are we doing there that slows down the presentation (probably something around the initialization).

Odrakir:
Fixes #591 

From my tests I think just by preventing `onCurrentCalendarDateChanged` from running twice when presenting the Calendar is enough. Please test if times are better for you too.


sofokli1:
i see a 25~30% improvement. better than nothing 😄

Odrakir:
Yeah, it's something. Please review.

PR #687, opened 2017-08-01
:bug: Fix mail compose not showing buttons sometimes

### LINKED ISSUE #682: Feedback form can miss Cancel and Send buttons
Possibly only the first time it is used after installing the app.

If this cannot be fixed, maybe it is time we look into a custom solution.

Odrakir:
Fixes #682 

Or at least hopefully, hard to say as it's not easy to reproduce.


sofokli1:
based on the docs looks more like something that changed in the past and most likely ios bug behaviour now days. will see what happens
```
/*
 The behavior of tintColor for bars has changed on iOS 7.0. It no longer affects the bar's background
 and behaves as described for the tintColor property added to UIView.
 To tint the bar's background, please use -barTintColor.
 */
@property(null_resettable, nonatomic,strong) UIColor *tintColor;
```

PR #691, opened 2017-08-02
Notification removal

### LINKED ISSUE #690: Disable Howdy! notifications
Shoo.

sofokli1:
closes #690


teferi/Services/Implementations/Interaction/NotificationSchedulingService.swift
```diff
@@ -25,14 +25,6 @@ class NotificationSchedulingService
         self.settingsService = settingsService
         self.smartGuessService = smartGuessService
         self.notificationService = notificationService
-        
-        locationService
```
Odrakir: You are not using the `locationService` anymore, why keep sending it in the constructor?

Odrakir: As a matter of fact... what does this class do if it doesn't send that notification? Because it seems to me that it does nothing at all. :fire:

sofokli1: maybe remove the whole notification service? 😛

Odrakir: No, we still send notifications for the user to enable HK.

sofokli1:
i have the feeling the PR went a bit further than the requirements of the issue

Odrakir:
Not at all, having old useless code is a recipe for disaster.

teferi/Models/NotificationType.swift
```diff
@@ -2,5 +2,4 @@ import Foundation
 
 enum NotificationType : String {
     case normal
-    case categorySelection
```
Odrakir: An enum with just one case? 🤔

teferi/AppDelegate.swift
```diff
@@ -129,16 +121,6 @@ class AppDelegate : UIResponder, UIApplicationDelegate
...
         
-        if #available(iOS 10.0, *) {
-            UNUserNotificationCenter.current().delegate = self
-        } else {
-            if let notification = launchOptions?[UIApplicationLaunchOptionsKey.localNotification] as? UILocalNotification {
-                didReceiveCategoryNotification = isCategorySelectionNotification(notification)
-            }
-        }
-        
-        appLifecycleService.publish(isInBackground ? .movedToBackground : .movedToForeground(fromNotification:didReceiveCategoryNotification))
```
Odrakir: We still need to know if the app moves to foreground or background. You can remove the `fromNotification` associated value though.

teferi/AppDelegate.swift
```diff
@@ -223,56 +205,15 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         pipeline.run()
         
         initializeWindowIfNeeded()
-     
-        notificationService.unscheduleAllNotifications(ofTypes: .categorySelection)
-        
-        appLifecycleService.publish(.movedToForeground(fromNotification:didReceiveCategoryNotification))
```
Odrakir: Same, we still need to know the the app moved to foreground.

teferi/Models/NotificationType.swift
sofokli1: i know it looks odd but since we are adding new notifications in the next mile stone i am keeping it for now.

Issue #549, opened 2017-05-03
Make every build have a different build number


Odrakir:
Probably with a script that counts the number of commits.

amulware:
@Odrakir Please resolve or close this.

If you're gonna work on it, please discuss the problem of multiple branches having the same numbers of commits. Either find a different solution or argue that it is not a problem we need to worry about.

Odrakir:
My argument agains that being a problem we should worry about is that we are only going to run into conflicts when building for distribution or beta testing, and as this is done always from the same two branches (develop and master) they'll always be increasing.

The only other time we do a test release from another branch is when force pushing to `test-build` but we do that sparingly enough that it shouldn't cause problems.

amulware:
Fair enough. If you can make it do this automatically on every build, go ahead. Just as long as it is fool proof and sets the build number in all places necessary of course.

PR #692, opened 2017-08-02
:page_with_curl: Set build number automatically with script

Odrakir:
Gets the build number from the number of commits and does this for every target.


teferiScripts/bump.rb
```diff
@@ -51,17 +41,15 @@ def bumpMajor
...
 
 verb = ARGV[0]
-validCommands = [ "build", "hotfix", "minor", "major"]
+validCommands = ["hotfix", "minor", "major"]
 
 commandDescriptions = {
-	"build" => "increments build number",
 	"hotfix" => "increments maintenance version and build number",
 	"minor" => "increments minor version and build number",
 	"major" => "increments major version and build number"
```
amulware: Fix descriptions please (and maybe include in minor and major that lower numbers are reset to 0 as well.

Odrakir: This doesn't change anything when you run the app in Xcode (simulator or device). It just sets the build number in the executable or when archiving, but nothing changes in code of config files, so nothing to commit.

I'll change the descriptions, but I don't understand what you mean by "and maybe include in minor and major that lower numbers are reset to 0 as well."

amulware: Unless the script was changed, when you increase the minor, it will reset the hotfix number. I.e.: 0.1.1 -> 0.2.0

sofokli1:
@Odrakir shouldn't this close #549 ?

Odrakir:
Correct

PR #693, opened 2017-08-02
:bug: Fix health permission not showing after notification

sofokli1:
Doesn't close any issue. But is known issue. The presentation of the permission was based on the install date which is set after the on-boarding is completed, but the notification was scheduled based on the first time the app was started. So the delay in the presentation was equal to the difference from when the app first started and the user finished the on-boarding.


amulware:
Good catch! I'll let support know that this won't be an issue coming next update. :)

PR #706, opened 2017-08-08
Force crash

### LINKED ISSUE #705: Secret way to force crash the app is not working anymore
It used to work, but somehow we lost it.

Odrakir:
Fixes #705


amulware:
This appears to make two changes to me.

PR #716, opened 2017-08-09
:sparkles: Log more meaningful data to fabric

### LINKED ISSUE #699: Average time slot editing per day per user
At the moment we can get a number of time slots edited per day per user but we would like to know out of how many so we can get the percentage.

sofokli1:
Closes #699 

Here we are logging:
- Which categories are created from the algo and manually from the users. (Also smart guesses)
- Which categories are edited (what it was and what is the new one) 
- Durations of the slots
- Locale `regionCode` (ex: US, GR) not based on location just phone's locale (is meant to be more of a hint)
- `Hour of the day` so that we have a better understating when the app is most used during the day (either opened to be viewed or opened to edit slots)
- `Day of the week` so that we have a better idea which day the app is used more


teferi/Pipeline/PersistencySink.swift
```diff
@@ -40,13 +43,23 @@ class PersistencySink : Sink
...
             else
             {
                 addedTimeSlot = timeSlotService.addTimeSlot(withStartTime: temporaryTimeSlot.start,
                                                                  category: temporaryTimeSlot.category,
                                                                  categoryWasSetByUser: false,
                                                                  location: temporaryTimeSlot.location?.toCLLocation())
+                
+                if let addedTimeSlot = addedTimeSlot
+                {
+                    metricsService.log(event: .timeSlotCreated(date: timeService.now, category: addedTimeSlot.category, duration: addedTimeSlot.duration))
```
amulware: In every other case of creation, we log the general event and a more specific one. Would it not be better to have a more specific event here as well, so we can easier see for example the ratio of smart guessed to not smart guessed?

PR #719, opened 2017-08-16
Timeline vote

### LINKED ISSUE #694: Ask user to upvote/downvote the timeline
Ask user to rate the timeline quality (how accurate was tracking and category guessing) by giving thumbs up or thumbs down on every evening.
- Show upvote/downvote option at the bottom of the timeline.
- User gets notified about the option once per day via notification. (let's send notification at 6pm)
- Notification copy: '**Rate your experience** - Did the app track your time accurately?'
- Tapping on a notification opens the app and scrolls user at the bottom of the timeline.
- Don't show this notification on Sundays. Voting is still available on Sundays.
- User can't change upvote/downvote after being already voted.
- Show 'Thank you' note after voting. Remove 'Thank you' note when app goes in the background or when user closes the app.
- User can rate days in the past as well.
- User can't turn this option off.
- Limit voting to last 7 days.

sofokli1:
Closes #694


teferi/en.lproj/Localizable.strings
```diff
@@ -25,6 +25,12 @@
 "WelcomeMessageTitle" = "That's it! 🎉";
 "WelcomeMessageSubTitle" = "Check back later to see where your time goes.";
 
+"VotingNotificationTittle" = "Rate your experience";
+"VotingNotificationMessage" = "Did the app track your time accurately?";
```
amulware: Up to @pe0ny, but maybe change the copy to 'Did Superday track your time accurately?'

pe0ny: +1 to change the copy to 'Did Superday track your time accurately?'

teferi/AppDelegate.swift
```diff
@@ -133,6 +133,8 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         
         initializeWindowIfNeeded()
         
+        notificationService.scheduleVotingNotifications()
```
sofokli1: This only needs to happen once. if called again it will override the previously scheduled notifications with same id. I didn't want to complicate the logic by adding extra checks and flags if it was already called.

teferi/Models/NotificationType.swift
```diff
@@ -3,4 +3,5 @@ import Foundation
 enum NotificationType : String
 {
     case normal
+    case repeatWeekly
```
sofokli1: no because we schedule 1 per every week day (except for Sunday) (total of 6) and make each of those repeatable for that week day every week.

teferi/Services/Implementations/Persistency/DefaultSettingsService.swift
```diff
@@ -165,9 +173,50 @@ class DefaultSettingsService : SettingsService
...
+    func setWelcomeMessageShown()
+    {
+        set(true, forKey: welcomeMessageShownKey)
+    }
+    func setWelcomeMessage(visible: Bool)
+    {
+        set(visible, forKey: welcomeMessageVisibleKey)
+    }
+    
+    func canShowVotingView(forDate date: Date) -> Bool
```
Odrakir: I think this kind of logic should be in the viewModel and not here, or at least some of it. This is not really a "setting", right?

teferi/Services/Implementations/Persistency/DefaultSettingsService.swift
```diff
@@ -165,9 +173,50 @@ class DefaultSettingsService : SettingsService
...
+    }
+    
+    func didVote(forDate date: Date)
+    {
+        var history = cleanedUpVotingHistory()
+        history.append(date.ignoreTimeComponents())
+        UserDefaults.standard.setValue(history, forKey: votingHistoryKey)
+    }
+    
+    private func cleanedUpVotingHistory() -> [Date]
```
Odrakir: Why "cleanedUpVotingHistory" why not "lastSevenDaysOfVotingHistory"?

teferi/Services/Implementations/Persistency/DefaultSettingsService.swift
```diff
@@ -165,9 +173,50 @@ class DefaultSettingsService : SettingsService
...
+            ( timeService.now.ignoreTimeComponents() == date.ignoreTimeComponents() ? timeService.now.hour >= Constants.hourToShowDailyVotingUI : true ) &&
+            !welcomeMessageVisible
+        else { return false }
+        
+        let alreadyVoted = !cleanedUpVotingHistory().contains(date.ignoreTimeComponents())
+        
+        return alreadyVoted
+    }
+    
+    func didVote(forDate date: Date)
```
Odrakir: This should be called "setVote" or "vote", as "didVote" is more common used for something that returns a Bool, first time I read it, that was what I expected it to do.

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -41,12 +41,14 @@ class MainViewModel : RxViewModel
             .map { [unowned self] () -> Bool in
                 guard self.timeService.now.ignoreTimeComponents() == self.settingsService.installDate!.ignoreTimeComponents()
                 else {
-                    self.settingsService.setWelcomeMessageHidden()
+                    self.settingsService.setWelcomeMessage(visible: false)
```
Odrakir: As mentioned before, wether the welcome message is visible or not at this very moment, shouldn't be part of the settings, it should be part of the view model.

sofokli1: How would you suggest i share this information with other viewmodels?

Odrakir: Which information?

sofokli1: if the welcome message is visible

Odrakir: Ok, you mean for the Timeline to know if the welcome message is there or not. I'm not sure about that, maybe somehow through the Pager, I don't want to overcomplicate things, but I think it's worth it to give it a thought.

In mi opinion using the Settings for something that is not a setting (more of a part of the "state" of the app) seems like a bad idea... But we can ask for a second opinion if you don't agree.

sofokli1: it is not that i do not agree, it's just that this seemed the easiest way to do it without overcomplicating it. Any idea to overcome this is welcome

sofokli1: This information has to travel from `Main` to `Pager` to `Timeline`and is especially useless for the `Pager` where it only passes through. Would it make sense to make a whole service just for this information? Or better leave it as is?

sofokli1: @Odrakir since the the informations is needed only at `Timeline` i am ok with the current logic of this PR. But would go with the new `Service` logic if an other component of the app needs the same information. what do you think?

Odrakir: Mmmm... I don't think having a new Service is a good solution either. In the end everything is a Service in this app...

If you think about it... there's always only one timeline visible at a time, so maybe it makes sense exposing that timeline (or some attribute of it) through the pager to the main view.

But, anyway... Is this all ever really needed? Are we ever going to show the timeline voting at the same time as the welcome message? It seems to me that's something which is impossible to happen, as the welcome message only shows the first time you log in and the voting only shows after 18 hours, isn't it?

In that case I would remove the stuff about the welcome message being visible completely from settings.

sofokli1: if you install the app after 18:00 the issue shows itself

sofokli1: it is a rare case but it is still a case

Odrakir: Oh, I'm sorry, I misunderstood, I thought it was 18 hours after install.

But then again, it makes no sense to show the voting if the user installs after 18:00 and all he sees is the default Leisure slot. So I propose two options here, whatever you prefer:
- Only show it if there are more than "n" slots (n could be 1 or 2 or...)
- Only show it after x hours after install.

I think probably the first one is easier. And then we can remove the messageVisible stuff.

sofokli1: i like the `"n" slots` option

Issue #704, opened 2017-08-07
Changing mid-night logic to take into account real sleep time


amulware:
The idea being to start the day when the user goes to sleep, not at midnight.

Apart from significant technical challenges, there are logical ones as well. Issues that have to be dealt with include:

- how to handle if a user has no sleep slot for a day? two days? two weeks?
- how to handle if a user has only small sleep slots spread across the day/night?
- how to handle that days can have basically arbitrary lengths? how will this work in statistics?
- how to handle a user whose week is into 6 waking/sleeping periods? does the week have 6 days? do they have fewer weeks/days that year?

And I'm sure there are other problems with this.

Odrakir:
I say those are a lot of problems. Now I understand Heidi's concerns, but for me they are not bad enough to get into this mess. 

A day is a day: a 24 hour period staring at 12:00 and we count how many hours you slept in that period. 

The thing would be totally different if we were making a sleep tracker in which, then yes, I would like to know the amount of time I slept "last night" where last night is the amount of time since I went to bed yesterday till I woke up today.

amulware:
While I'd like this, I don't see it being done anytime soon.

PR #720, opened 2017-08-18
Midnight cap pipe

### LINKED ISSUE #695: Fill gap between midnight and first timeslot of the day
Of course only if there is one. I think in most cases it makes sense to set this to `sleep` automatically, but I'm not sure what to do with people that stay up till 1am or something.

Ideas/suggestions welcome.

In either case, it would be nice if we could reach a 24h coverage of each day, just so that no time is 'lost' when looking at statistics.

sofokli1:
Closes #695


Odrakir:
I would like to discuss this a bit further before approving. It seems to me a little risky to split slots like that, I know that right now probably it's ok, but it doesn't seem like the best solution to me.

First, I think this doesn't address @pe0ny's concerns about the sleep being split in half, right? What Heidi wanted was for the sleep to happen all at the end of a day, and the next day starts when the user wakes up.

And secondly, I was thinking: As what we are really looking for is a way to fix what the user sees... Why don't we just fix them when showing them to the user? So, if he sleeps from 10 PM to 8 AM next day, we store that big slot in the DB, but when we transform those slots into timeline items, we decide if we need to split that, cap them or directly not show them because they belong to the previous/next day.

Do I make sense? @amulware @sofokli1 @pe0ny

amulware:
We decided to not do any fancy shenanigans with starting the day before/after sleeping, since there are way too many edge cases to handle. Discussion can continue here: #704

With regards to only making this change when showing data to the user: I think doing it at generation is easier. Otherwise, when fetching data for a day, we have to fetch one additional time entry for the day before, which seems strange to me and complicates the query.

Odrakir:
Ok then. If we are going to split the slot, I agree it's not worth it to do it any other way. If we ever do the other thing, we can discuss this again 😀

PR #723, opened 2017-08-30
:construction: Scroll to correct day when opening from notification

sofokli1:
Fixes problem when opening daily voting notification for past day. The app was not scrolling to that day.


sofokli1:
@Odrakir not sure why the test that i changed in `758d1d5` was failing. let me know if it makes sense.

teferi/Services/Protocols/State/AppLifecycleService.swift
```diff
@@ -18,4 +18,17 @@ extension AppLifecycleService
             }
             .mapTo(())
     }
+    
+    var startedOnDailyVotingNotificationDateObservable : Observable<Date>
+    {
+        return self.lifecycleEventObservable
+            .filter {
```
Odrakir: I think you don't need this first filter.

PR #736, opened 2017-09-01
Introduce user feedback and metrics

sofokli1:
## User facing:
- Split slots that go past midnight into two slots divided at midnight
- Add Timeline daily voting component + daily notifications (except for Sunday)
- Add weekly rating component + weekly notification (only on Sunday)
- Clears Notification Canter every time the app is opened

## Bug fixes
- Fix crash in weekly summary

##  Internal / Refactoring
- Improve fabric logging with more data from user input in daily voting and weekly review
- Add customisable padding for modal popup VCs
- Make pie chart white circle size customisable


amulware:
Changelog isn't very user-friendly, and there is a typo in the first line (slot/slots). I can rewrite it next week if you like.

sofokli1:
If it is not too much bother

PR #740, opened 2017-09-04
Fix not scrolling to bottom for voting

### LINKED ISSUE #738: Timeline vote notification doesn't direct user to the right place
Timeline vote notifications work like a charm but just open the correct timeline, though these should scroll to the 'vote' section, so user can give thumbs up/down right away. At the moment these are just opening the timeline at the top.

sofokli1:
closes #738


teferi/AppDelegate.swift
```diff
@@ -132,8 +132,6 @@ class AppDelegate : UIResponder, UIApplicationDelegate
             }
         }
         
-        appLifecycleService.publish(isInBackground ? .movedToBackground : .movedToForeground(withDailyVotingNotificationDate: dailyVotingNotificationDate))
```
Odrakir: Why is this not needed anymore?

teferi/Services/Implementations/State/DefaultAppLifecycleService.swift
```diff
@@ -3,7 +3,7 @@ import RxSwift
 class DefaultAppLifecycleService : AppLifecycleService
 {
     let lifecycleEventObservable : Observable<LifecycleEvent>
-    private let lifecycleSubject = PublishSubject<LifecycleEvent>()
+    private let lifecycleSubject = BehaviorSubject<LifecycleEvent>(value: LifecycleEvent.movedToForeground(withDailyVotingNotificationDate: nil))
```
Odrakir: How about it being a `BehaviorSubject<LifecycleEvent?>` with a nil initial value and filtering out nil in the `lifecycleEventObservable`? Otherwise the first event will be moved to foreground without notification date.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -57,6 +57,10 @@ class TimelineViewController : UIViewController
     {
         super.viewDidLoad()
         
+        viewModel.dailyVotingNotificationObservable
```
Odrakir: Can you move this down to where the other viewModel bindings are? Maybe it's even worth it to create a `bindViewModel` method for all of them, your call.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -125,21 +129,19 @@ class TimelineViewController : UIViewController
             self.delegate?.didScroll(oldOffset: old + topInset, newOffset: new + topInset)
         })
         .addDisposableTo(disposeBag)
+        
+        viewModel.didBecomeActiveObservable
```
Odrakir: Won't this show the voting UI everytime the app comes from background on that day, regardless if you voted before or not?

teferi/AppDelegate.swift
sofokli1: my bad, i thought it was the same as the one in `applicationDidBecomeActive`

Issue #737, opened 2017-09-04
Weekly user review crashes


pe0ny:
After getting notification and tapping on it, it tried to open the app and crashed before the timeline appeared.
I tried to open the app again, but it didn't show me any popup (though it should have when user haven't rated it yet), just regular timeline.

Can't say what happens when i open the app during the time when it should display the popup, with out interacting with the notification.

sofokli1:
As far as i can tell based on Crashlytics the crash is not related to the weekly rating . Seems related to the pager for the timeline. Regarding the weekly rating not showing again there is a PR waiting for review for that.

sofokli1:
@pe0ny if you don't see this happening, please close

PR #741, opened 2017-09-04
Fix weekly rating

sofokli1:
This PR should  fix the issue where the rating view was opening every time the app opens during the allowed period. Was not using settingService to keep track of the last time that it was presented.

Partially helps with #737


amulware:
Should this not be merged into the release branch?

teferi/Services/Implementations/State/DefaultAppLifecycleService.swift
```diff
@@ -3,7 +3,7 @@ import RxSwift
 class DefaultAppLifecycleService : AppLifecycleService
 {
     let lifecycleEventObservable : Observable<LifecycleEvent>
-    private let lifecycleSubject = PublishSubject<LifecycleEvent>()
+    private let lifecycleSubject = BehaviorSubject<LifecycleEvent>(value: LifecycleEvent.movedToForeground(withDailyVotingNotificationDate: nil))
```
Odrakir: I've commented this change in another PR (why is it there twice?). I think LifecycleEvent should be optional, filtering it out in the Observable.

PR #745, opened 2017-09-11
:bug: Fix where settingsService is used

sofokli1:
Fix where settingsService is used


Odrakir:
I don't see how this PR would change anything... I think the problem must be somewhere else.

@pe0ny , just to clarify: The app needs to be installed for 7 days or more for this to appear. You are not deleting the app to install the new build, right?

pe0ny:
@Odrakir Yes. I installed the app on Aug 29th and haven't deleted it since.

sofokli1:
@Odrakir This change still makes sense to be merged

Odrakir:
Care to explain why? I don't see much of a difference...

sofokli1:
It makes more sense to set the setting when we actually load the VC and not when we ask if we can load the VC

sofokli1:
@Odrakir Is your question answered?

PR #759, opened 2017-09-12
Changes onboarding for CoreMotion

### LINKED ISSUE #725: Rework onboarding for CM
Ask user to give access for location and CoreMotion during the onboarding flow.
Don't allow user to proceed and use the app without.

Odrakir:
Closes #725


teferi/Services/Implementations/Tracking/DefaultMotionService.swift
```diff
@@ -0,0 +1,99 @@
...
+
+fileprivate extension MotionEvent
+{
+    static func from(activity: CMMotionActivity) -> MotionEvent
+    {
+        var type: MotionEventType = .other
+        if activity.automotive {
+            type = .auto
+        } else {
+            if activity.running {
```
sofokli1: Use a `switch` without default value so that we do not miss anything now and in the future.

teferi/Services/Implementations/Tracking/DefaultMotionService.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    func getActivities(since start: Date, until end: Date) -> Observable<[MotionEvent]>
+    {
+        guard CMMotionActivityManager.isActivityAvailable() else {
+            return Observable.error(MotionServiceError.notAvailable)
+        }
+        
+        return Observable.create { [unowned self] observer in
+            
+            self.motionActivityManager.queryActivityStarting(from: start, to: end, to: OperationQueue.main, withHandler: { (activities, error) in
```
sofokli1: why `OperationQueue.main`?
wouldn't it make more sense to run the handler on a background thread and execute on main thread only
```
observer.onNext(events)
observer.onCompleted()
```

teferi/Models/MotionEvent.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+
+enum MotionEventType: String
+{
+    case walk = "walk"
```
sofokli1: remove the  `= "..."` because it is not needed when enum is a string. it's implied. Correct me if i am wrong.

teferi/Models/MotionEvent.swift
```diff
@@ -0,0 +1,51 @@
...
+    case cycling = "cycling"
+    case auto = "auto"
+    case other = "other"
+    case still = "still"
+}
+
+struct MotionEvent: Equatable
+{
+    let timeStamp: Date
+    let end: Date
```
sofokli1: why is one named `timestamp` and the other `end`. what do these 2 properties mean?

teferi/Models/MotionEvent.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
```
sofokli1: A lot of constructs in this file but ok

teferi/Services/Implementations/Tracking/DefaultMotionService.swift
Odrakir: You are right about not using the main thread. I'm going to use the current queue, so the caller of this method can decide in an Rx manner which queue to use.

teferi/Services/Implementations/Tracking/DefaultMotionService.swift
Odrakir: It can't be a switch, because those are different properties. This is a little ugly, but it's the best I could come with...

Podfile.lock
```diff
@@ -15,7 +15,7 @@ PODS:
   - SnapKit (3.2.0)
   - SwiftGen (4.1.0)
   - SwiftLint (0.16.1)
-  - SwiftyBeaver (1.4.0)
+  - SwiftyBeaver (1.4.1)
```
sofokli1: why is this in this commit?
pod file still says only `pod 'SwiftyBeaver'` so it can't build out of the box

teferi/Services/Implementations/Tracking/DefaultMotionService.swift
```diff
@@ -25,7 +25,7 @@ class DefaultMotionService: MotionService
         
         return Observable.create { [unowned self] observer in
             
-            self.motionActivityManager.queryActivityStarting(from: start, to: end, to: OperationQueue.main, withHandler: { (activities, error) in
+            self.motionActivityManager.queryActivityStarting(from: start, to: end, to: OperationQueue.current ?? OperationQueue(), withHandler: { (activities, error) in
```
sofokli1: Tested and this still runs on main, in a way it makes sense since the function is run main thread aka. `.current`
If it is not main (which it is) the `on next` and `on complete` will run on that thread even though the caller is most likely on main which would mean interesting crashes when updating UI

Podfile.lock
Odrakir: I guess a new version of that library was released while doing this. The pods where outdated in this branch, so I had to do `pod install` and as we are not indicating a specific version for this pod, it updated it self.

teferi/Services/Implementations/Tracking/DefaultMotionService.swift
Odrakir: That's why it's the caller who decides which thread to use for that and in which thread to return, via the Rx operators `subscribeOn` and `observeOn`

PR #765, opened 2017-09-13
Fix iOS 11 location permissions

### LINKED ISSUE #712: Fix location permission onboarding
It appears that currently, it's not working at all.

We also most likely have to reposition the pointing emoji to point at the correct location for the updated permission notification (and this has to be different depending on which version the OS is).

Odrakir:
Closes #712 and also changes the location permission's text.


sofokli1:
Fix the test issue and i will approve

Odrakir:
My bad there WAS a problem with tests.

PR #771, opened 2017-09-13
:triangular_ruler::construction: Fix navigationBar buttons for iOS 11

### LINKED ISSUE #760: Bar buttons don't seem to be working in iOS 11
Tried in the simulator and they don't do anything. I didn't investigate any further.

sofokli1:
Closes #760


sofokli1:
negative fixed space don't work anymore. not sure about UIEdgeInsets since we add uiBarButton items now not subviews directly. Also i talke with @pe0ny and shi was ok with the change in space between the buttons.

pe0ny:
IMO it looked good yep 👍

Odrakir:
If she's ok, then you are good to go.

PR #775, opened 2017-09-14
Bug fixes

### LINKED ISSUE #773: Notification for weekly rating is fired even in the first week of install
Notification will fire for the first sunday in the week of the install but the UI will not show.

@pe0ny is it possible you experienced something like this?

sofokli1:
Closes #773 #772 
Alos fixes race condition when unscheduling and scheduling local notifications.
Also does not show voting UI and notification for first day of install.


teferi/Services/Implementations/Interaction/PostiOSTenNotificationService.swift
```diff
@@ -64,33 +66,43 @@ class PostiOSTenNotificationService : NotificationService
...
-                    {
-                        self.notificationCenter.removePendingNotificationRequests(withIdentifiers: [request.identifier])
-                    }
+            
+            self.notificationCenter.getPendingNotificationRequests { (requests) in
+                requests.forEach({ (request) in
+                    givenTypes.forEach({ (type) in
+                        if request.identifier.contains(type.rawValue)
+                        {
+                            self.notificationCenter.removePendingNotificationRequests(withIdentifiers: [request.identifier])
```
Odrakir: Instead of looping through all notifications and calling this method for the selected ones... Wouldn't it be easier to read and more efficient to filter and map notifications to the appropiate identifiers and call this method just once?

```swift
let identifiers = notifications.filter(...MATCHING_TYPE...).map(...TO_IDENTIFIER...)
self.notificationCenter.removeDeliveredNotifications(withIdentifiers: identifiers)
```

PR #782, opened 2017-09-21
:construction: Changes on On boarding for consistency

Odrakir:
Basically I'm changing this so location and motion services APIs behave the same, even if the underlaying SDKs are different.


teferi/AppDelegate.swift
```diff
@@ -111,9 +111,7 @@ class AppDelegate : UIResponder, UIApplicationDelegate
                 dailyVotingNotificationDate = dailyVotingDate(notification)
             }
         }
-        
-        appLifecycleService.publish(isInBackground ? .movedToBackground : .movedToForeground(withDailyVotingNotificationDate: dailyVotingNotificationDate))
```
sofokli1: why remove this?

Odrakir: Apparently we are publishing `.movedToForeground` twice, once here and another time in `applicationDidBecomeActive`...

But now that you point this out... maybe it has to stay there but only for `.movedToBackground`. What do you think?

PR #785, opened 2017-09-25
Core motion permissions for current user

### LINKED ISSUE #774: Ask permission to motion data from existing users
We need to ask access to motion data from existing users as well.

@Odrakir What are our options with HK - it would be nice to say to users we won't need access to their HK data anymore (because we don't) but is it actually possible to remove/disable HK access?

sofokli1:
Closes #774


Odrakir:
Can you first fix conflicts? This will probably change things, as we are touching the same stuff.

Odrakir:
It doesn't quite work yet, in part because of my core motion code. I've got an idea to fix this once and for all, but it'll only work once we merge my PR, let's do that and then we can discuss this.

Issue #788, opened 2017-09-28
UI issue: Last onboarding screen 


pe0ny:
After installing the app and going through onboarding screens (allowing both: location and CM), UI shows empty state + last onobarding step at the same time. When i swipe left or right, i can see the correct screen:

<img width="763" alt="screen shot 2017-09-28 at 15 58 06" src="https://user-images.githubusercontent.com/5064445/30967742-d788e026-a465-11e7-862f-9b8005c95e96.png">

PR #793, opened 2017-09-29
Motion permissions

### LINKED ISSUE #754: CM disabled screen
Make sure to show this screen when:
a) User didn't allow access to CM during the onboarding flow
b) Has turned off CM from the settings
Don't allow user to proceed and use the app without enabling CM.

Make sure it won't overlap with #753 
When user has both disabled (location + CM), show location one first and then CM

timeline cm disabled 1x

Odrakir:
Closes #754 & #788 

As we are updating the settings when we find out the motion permissions are disabled, I use that to keep the UI updated (with the `motionPermissionGranted` Observable).

Also, in the case of the timeline dialog, iOS always restarts the app when the user changes the motion permission from settings, so that's why behavior changes a little there. I always set that setting to true, when the MainViewModel loads, as we'll no if it's really true when trying to generate the timeline.


teferi/TimelineGenerator/TimelineGenerator.swift
```diff
@@ -33,6 +33,8 @@ class TimelineGenerator
     
     func execute() -> Observable<Void>
     {
+        print("EXECUTE")
```
sofokli1: remove print

teferi/UI/Common/Views/LoadingView.swift
```diff
@@ -51,6 +49,7 @@ class LoadingView: UIView
     
     func show()
     {
+        print("Show")
```
sofokli1: remove print

teferi/UI/Common/Views/LoadingView.swift
```diff
@@ -82,6 +81,7 @@ class LoadingView: UIView
     
     func hide()
     {
+        print("Hide")
```
sofokli1: remove print

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -157,7 +160,7 @@ class MainViewModel : RxViewModel
 
         super.init()
         
-        appLifecycleService.movedToForegroundObservable
+        appLifecycleService.movedToForegroundObservable.debug()
```
sofokli1: remove debug

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -166,10 +169,12 @@ class MainViewModel : RxViewModel
                 return locationService.currentLocation
                     .subscribeOn(OperationQueueScheduler(operationQueue: OperationQueue()))
                     .do(onNext: settingsService.setLastLocation)
+                    .debug()
```
sofokli1: remove debug

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -166,10 +169,12 @@ class MainViewModel : RxViewModel
...
                     .trackActivity(self.locatingActivity)
             }
             .flatMap { [unowned self] _ in
                 return self.timelineGenerator.execute()
+                    .debug()
```
sofokli1: remove debug

teferi/UI/Modules/On Boarding/OnboardingPage4.swift
```diff
@@ -14,8 +14,15 @@ class OnboardingPage4 : OnboardingPage
     {
         viewModel.requestCoreMotionAuthorization()
         
-        viewModel.motionAuthorizationChangedObservable
+        viewModel.motionAuthorizationChangedObservable.debug()
```
sofokli1: remove debug

PR #794, opened 2017-09-29
Change to manual profiles

Odrakir:
This is to fix code signing in bitrise


Odrakir:
I do, but Bitrise doesn't.

PR #833, opened 2017-10-23
Super timeline

sofokli1:
- Fix provisioning profiles
- Fix movement slot margins and alignment in timeline
- Simplify smartGuesses to use previews timeSlots from past 2 weeks and just use location for calculations
- Ask user permission for notifications
- Rename `Auto` to `Transport`
- Add slot detail view (category edit, maps etc.)
- Fix bug for long height on slots with small duration
- Add new movement slots in timeline
- Improve timeSlot line height calculation based on duration


amulware:
User-facing changelog:
```
- Better visualization for movement slots
- Distinguishing walk, run, cycle and transport
- New editing UI
- Various fixes and improvements
```

PR #839, opened 2017-10-24
Workaround iOS 10 UI glitch

sofokli1:
Fixes the blur background in edit timeSlot view in iOS 10 where it looked completely transparent.


amulware:
I believe the build number has to be updated as well?

amulware:
Oh wait.. I'm thinking out of date, we automated that.

PR #856, opened 2017-11-07
New goal popup

### LINKED ISSUE #825: New goal popup
To set a goal user have to pick:
- how much time he/she wants to spend on it (30min, 45min, 1h, 1.5h, 2h, 3, 4, 5, 6, 7, 8, 9, 10h)
- on what he/she wants to spend time (pick a category)

Other notes:
- user can slide durations and categories from left to right and from right to left (similar how current editing categories works)
- tapping 'Set a goal' button takes users to Goals view

goal popup 1x

sofokli1:
Closes #825


teferi/UI/Common/TimeIntervalHelper.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
```
Odrakir: This file shouldn't be part of this PR.

teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
```diff
@@ -67,16 +67,21 @@ class GoalViewModel
     //MARK: Private Methods
     private func withMissingDateGoals(_ goals: [Goal]) -> [Goal]
     {
+        guard goals.count > 0 else { return [] }
```
Odrakir: This method is getting too complicated for someone as dumb as me to understand. Wouldn't it be easier to just get the earliest goal and loop through all dates until present checking if that date is an array of goals? Just a suggestion...

teferi/UI/Modules/Goals/NewGoal/Views/HorizontalPicker.swift
```diff
@@ -0,0 +1,300 @@
+import UIKit
+import RxSwift
+
+class HorizontalPicker<ItemType: Equatable> : UIControl, TrigonometryHelper, UIDynamicAnimatorDelegate
```
Odrakir: I know this probably comes from the time slot edit view (it was done before I joined Toggl). But... why do your own implementation instead of just using a `UICollectionView`? I think a solution with much less code and even with more beautiful result could be done by using collection views instead.

sofokli1: couldn't estimate how long it would have taken to do it with a collection view. reusing the same logic from other component seemed faster.

PR #857, opened 2017-11-08
:bug: Try to continue last stored slot if possible

### LINKED ISSUE #821: App creates new slot almost right after creating manual slot
> when i create a manual slot the app then starts a new slot after that usually.  eg. i sit down, open the app, create a manual entry for work and put down the phone. i'd then expect the app to keep tracking that entry but instead most often it creates a new blank empty right after that (i assume since it notices i stopped moving around) so whenever i create a manual entry it's a 0min entry followed immediately by an uncategorized one that's the actual time i spent there.

Odrakir:
Closes #821 

I'm checking wether or not we should continue the last stored slot when persisting the temporary time slots. As we have the same problem with automatic generated TS I add that check too.


Odrakir:
I think we are still within reasonable usage (could be moved, but I'm just not sure is worth it for now)

PR #865, opened 2017-11-13
:bug: Fix modal detail presentation view

### LINKED ISSUE #859: View goes black when changing tabs
If I show a detail time slot view, change to another tab without closing it, and go back, the view is black.

Even if we fix this, which shouldn't be hard, I think it makes sense for the whole popup to appear over the tab bar, so this is just not possible.

img_3905

Odrakir:
Closes #859 

The modal presentation had to be `.custom` so it appears over the tabBar. That implied also some customization in `ModalPresentationController`


teferi/UI/Common/PresentationControllers/ModalPresentationController.swift
```diff
@@ -53,6 +61,8 @@ class ModalPresentationController: UIPresentationController
     
     override func presentationTransitionWillBegin()
     {
+        guard let dimmingView = dimmingView else { return }
+        
         containerView?.insertSubview(dimmingView, at: 0)
```
georgmay: You could use `!` in here, as explained in here:
https://developer.apple.com/documentation/uikit/uipresentationcontroller/1618332-containerview

This view is expected to be created on time when transition occurs, otherwise it'll be undefined behaviour anyway.

Odrakir: It would crash in the next line any way, so yes, I agree with you.

Odrakir:
None of those changes are part of this commit, but I completely agree with you, I'll fix them and submit again.

PR #874, opened 2017-11-13
Results notification

### LINKED ISSUE #830: Inform the user about the results on the next day
If user has not yet seen his/her results, send a notification on the next day saying: 'Results are in for September 24. Check it out'.

goals - notification v1 1x

Odrakir:
Closes #830 

This adds a new type of notification, I also did some cleanup. I'm guessing weekly notifications also make more sense to be setup when going to the background instead of where they are now, but we can see about that.


teferi/Services/Implementations/Interaction/DefaultNotificationService.swift
```diff
@@ -141,6 +139,22 @@ class DefaultNotificationService : NotificationService
...
+            goal.date.ignoreTimeComponents() == timeService.now.ignoreTimeComponents()
+            else { return }
+        
+        let tomorrowAt11 = timeService.now.tomorrow.setHour(11)
+        
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "MMMM dd"
+        let todayDateString = dateFormatter.string(from: timeService.now)
+        
+        let message = L10n.resultsNotificationMessage.replacingOccurrences(of: "[date]", with: todayDateString)
```
sofokli1: why not use string with format?

teferi/Services/Implementations/Interaction/PreiOSTenNotificationService.swift
```diff
@@ -0,0 +1,92 @@
+import Foundation
```
sofokli1: why add this file back?

Odrakir: ⁉️ 🤷‍♂️

PR #877, opened 2017-11-14
:bug: Fix placeholder goals filling bug

### LINKED ISSUE #871: App crashes when adding a goal if there's already one yesterday
I think it has something to do with the table views.

Odrakir:
Closes #871 

In the end there was an ugly bug in the code for filling in the placeholder goals. When there was a single goal for yesterday, and you try to add a goal for today, the app crashed because that method returned duplicate goals.

So I finally replace it with a much simpler and easier to read implementation.


teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
```diff
@@ -86,30 +86,28 @@ class GoalViewModel
...
-            goalsToReturn.append(goal)
-        }
+            date = date.add(days: 1)
+        } while !date.isSameDay(asDate: today.tomorrow)
         
-        return goalsToReturn
+        return goalsToReturn.reversed()
     }
     
     private func placeHolderGoals(fromDate: Date, toDate: Date) -> [Goal]
```
sofokli1: if you are not using this function you should remove it in this PR

PR #879, opened 2017-11-14
:bug: Make sure nav bar is visible in every view

### LINKED ISSUE #862: Make sure nav bar is visible in every view
- [x] Nav bar should be visible in all views: Timeline, Goals, Reports, Settings
- [x] Nav bar paddings from left and right should be: 16pt from left and 8pt from right
- [x] Superday text should be 12pt from the Superday icon
- [x] Show Calendar icon only in Timeline view (Not in Goals or Reports).
- [x] Calendar icon colour should be #B7B9BB

timeline 1x

georgmay:
Closes #862 

I've made a bit of reorganisation in the code related to navigation and so on, also fixed a couple of UI issues.


teferi/UI/Modules/Navigation/NavigationController.swift
```diff
@@ -4,6 +4,17 @@ import RxSwift
 import RxCocoa
 import Crashlytics
 
+fileprivate extension UIBarButtonItem
```
Odrakir: Move this extension either to the bottom of the file or to its own file (in the extensions group)

teferi/UI/Modules/Navigation/NavigationController.swift
```diff
@@ -84,30 +103,29 @@ class NavigationController: UINavigationController
...
     {
         super.pushViewController(viewController, animated: animated)
+        pushedViewController = viewController
         
-        setupNavigationBar(viewController: viewController)
+        setupNavigationBar(for: viewController)
     }
     
-    private func setupNavigationBar(viewController: UIViewController)
+    private func setupNavigationBar(for viewController: UIViewController)
```
Odrakir: This method ends up being called twice for every view controller. I would rather have the TimelineViewController add the calendar button itself to its navigation item, instead of having the NavController do it.

georgmay: That's tricky, as Timeline got to know about the layout of the buttons in NavBarItems, and the calendarButton exists in NavigationController, so there'll be 2 classes knowing about the same thing. 

I can add the method in navigationcontroller like `insertCalendarButton` and call it from TimelineViewController, wdyt.

Odrakir: But why having the calendarButton in the nav controller at all? I'm saying to move it all to the Timeline.

georgmay:
@Odrakir please re-review.

PR #880, opened 2017-11-15
New Goal: Replace custom solution with collection views

Odrakir:
In order to be able to use the same view for editing a goal the selectors need to start in the current selection offset. That forces this change, because it wasn't possible with the previous solution.

Basically this view let's us setup a new goal, selecting the category and the amount of time we want to reach. In a next PR this will be used to edit existing goals as well.

I'm subclassing `UICollectionView` for the times and categories selectors. There's some custom code to make it snap to the cells. Also, in the case of the categories one there's some code to make it loop. It's actually not an infinite loop, it repeats the fixed number of categories 150 times, and then uses the modulo (%) operator to ask the datasource for the appropiate model. I found this to be the easiest and less "dirty" way of faking a loop.


teferi/UI/Modules/Goals/NewGoal/Views/GoalTimeCell.swift
```diff
@@ -0,0 +1,26 @@
+//
```
georgmay: Delete header comment

teferi/UI/Modules/Goals/NewGoal/Views/GoalCategoryCell.swift
```diff
@@ -0,0 +1,43 @@
+//
```
georgmay: Delete header comment

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
```diff
@@ -0,0 +1,119 @@
...
+    
+    override func layoutSubviews()
+    {
+        super.layoutSubviews()
+        
+        let layout = collectionViewLayout as! UICollectionViewFlowLayout
+        let inset = frame.width / 2 - layout.itemSize.width / 2
+        contentInset = UIEdgeInsets(top: 0, left: inset, bottom: 0, right: inset)
+        
+        if !didLayout {
```
georgmay: Maybe better to move it to another function

teferi/UI/Modules/Goals/NewGoal/NewGoalViewController.swift
```diff
@@ -126,29 +69,45 @@ class NewGoalViewController: UIViewController
...
     }
     
-    private func viewForCategoryPickerItem(_ category: Category, viewSize: CGSize) -> UIView
+    func cell(at row: Int, for collectionView: UICollectionView) -> UICollectionViewCell
     {
-        let view = CategoryWithNameView.fromNib()
-        view.frame = CGRect(origin: .zero, size: viewSize)
-        view.category = category
-        return view
+        if collectionView == timesCollectionView {
```
georgmay: Code might look clearer if you'd create `TimeCollectionViewDatasource` and `CategoryCollectionViewDatasource`

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+    }
+}
+
+extension CustomCollectionView: UICollectionViewDataSource
+{
+    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int
+    {
+        return totalNumberOfItems * (loops ? 300 : 1)
```
georgmay: That's one for me, what is this `300` stands for?

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
```diff
@@ -0,0 +1,119 @@
...
+}
+
+protocol CustomCollectionViewDelegate
+{
+    func itemSelected(for collectionView: UICollectionView, at row: Int)
+}
+
+class CustomCollectionView: UICollectionView
+{
+    var loops: Bool = false
```
georgmay: I'd split implementation in 2 classes as such, because almost in every function you get this `if loops` check.

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
```diff
@@ -0,0 +1,119 @@
...
+        
+        let layout = collectionViewLayout as! UICollectionViewFlowLayout
+        let inset = frame.width / 2 - layout.itemSize.width / 2
+        contentInset = UIEdgeInsets(top: 0, left: inset, bottom: 0, right: inset)
+        
+        if !didLayout {
+            didLayout = true
+            if loops {
+                let cellWidth = layout.itemSize.width + layout.minimumInteritemSpacing
+                contentOffset = CGPoint(x: -contentInset.left + cellWidth * CGFloat(150 * totalNumberOfItems), y: 0)
```
georgmay: Why 150?

teferi/UI/Modules/Goals/NewGoal/NewGoalViewModel.swift
```diff
@@ -14,6 +14,14 @@ class NewGoalViewModel
         return self.categoryProvider.getAll(but: .unknown)
     }
     
+    typealias Minutes = Double
+    var goalTimes:[GoalTime] = {
+        let times: [Double] = [30, 45, 60, 1.5 * 60, 2 * 60, 3 * 60, 4 * 60, 5 * 60, 6 * 60, 7 * 60, 8 * 60, 9 * 60, 10 * 60]
```
georgmay: Is it something that gonna stay like that for a while, or is something that's gonna change a lot?

Odrakir: You never know...

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
Odrakir: The number of times the loop repeats. I'll try to clarify using a named property.

teferi/UI/Modules/Goals/NewGoal/Views/CustomCollectionView.swift
Odrakir: I see your point, but at this moment I think a new class would be too much. Let's leave it like that for now.

teferi/UI/Modules/Goals/NewGoal/NewGoalViewModel.swift
georgmay: It's just, it might be better to move somewhere else like config file or something.

teferi/UI/Modules/Goals/NewGoal/NewGoalViewController.swift
```diff
@@ -126,29 +83,45 @@ class NewGoalViewController: UIViewController
...
-            String(format: "%02d", (Int(duration) / 60) % 60) :
-            duration == 1.5*hourInSeconds ?
-                String(format: "%.1f", (duration / 3600)) :
-                String(format: "%.01d", (Int(duration) / 3600))
+        let category = viewModel.categories[row]
+        viewModel.categorySelectedVariable.value = category
+    }
+}
+/*
+extension NewGoalViewController: CustomCollectionViewDatasource
```
georgmay: Why keeping it in the code? I mean you could get it back from git history anytime..

Odrakir: Totally unintended.

Issue #887, opened 2017-11-16
Icon in goal header is not aligned correctly


Odrakir:
It's a little bit lower than it should. Check Zeplin.

![img_6bfeef53ada6-1](https://user-images.githubusercontent.com/724743/32878834-ead90954-caa7-11e7-96ff-fa01ee8ec966.jpeg)

PR #893, opened 2017-11-21
:construction: Add goal failed messages

### LINKED ISSUE #828: Goal failed view
- Wait until midnight before displaying failure in the app. (When user achieved the goal, show it immediately).
- User can either see his/her score
a) By checking his/her phone after midnight or after that
b) By notification if user didn't check his/her phone on the last day #830

sofokli1:
closes #828


teferi/UI/Modules/Goals/GoalList/Helper/GoalMessageProvider.swift
```diff
@@ -0,0 +1,138 @@
...
+        {
+            return L10n.goalHeaderTodayIWantToSpendHours(String(components.hour!))
+        }
+        else
+        {
+            return L10n.goalHeaderTodayIWantToSpendMinutes(String(components.minute!))
+        }
+    }
+    
+//    func message(forGoal goal: Goal) -> (message: String?, categoryHidden: Bool)
```
Odrakir: Commented code

teferi/UI/Modules/Goals/NewGoal/NewGoalViewController.swift
```diff
@@ -30,7 +30,7 @@ class NewGoalViewController: UIViewController
         titleLabel.text = L10n.newGoalTitle
         topDescriptionLabel.text = L10n.newGoalTopDescription
         middleDescriptionLabel.text = L10n.newGoalMiddleDescription
-        newGoalButton.setTitle(L10n.newGoalActionButtonTitle, for: .normal)
+        newGoalButton.setTitle(L10n.newGoalActionButtonTitleSetAGoal, for: .normal)
```
Odrakir: This wouldn't work, it always gets the same title, instead of the dynamic one. Also, this is overriden at the end of this function.

Odrakir: And it shouldn't be part of this PR.

teferi/UI/Modules/Goals/NewGoal/NewGoalViewModel.swift
```diff
@@ -40,9 +40,9 @@ class NewGoalViewModel
         if let goalToBeEdited = goalToBeEdited {
             initialCategory = goalToBeEdited.category
             initialTime = GoalTime(goalTime: goalToBeEdited.targetTime)
-            buttonTitle = "Done"
+            buttonTitle = L10n.newGoalActionButtonTitleDone
```
Odrakir: This shouldn't be in this PR, and also I'm not sure I get the necessity of doing this when we are not translating the app. Right now it just adds more points of possible failure.

teferi/UI/Modules/Goals/GoalList/Views/GoalHeader.swift
```diff
@@ -9,26 +9,11 @@ class GoalHeader: UIView
     @IBOutlet private weak var separatorView: UIView!
     @IBOutlet private weak var bottomConstraint: NSLayoutConstraint!
     
-    func configure(withGoal goal: Goal?, message: String?)
+    func configure(withGoal goal: Goal?, message: String?, categoryVisible: Bool)
```
Odrakir: Not sure what's going on with this configuration, but it's not working correctly. The button to add a new goal doesn't show when it should (when the fail message shows or even when it doesn't but the last goal has failed)

teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
```diff
@@ -74,11 +78,26 @@ class GoalViewModel
...
     }
     
-    func message(forGoal goal: Goal?) -> String?
+    func messageAndCategoryVisibility(forGoal goal: Goal?) -> (message: String?, categoryVisible: Bool)
     {
-        guard let goal = goal else { return nil }
+        guard let goal = goal else { return (goalHeaderMessageProvider.newGoalMessage, false) }
         
-        return goalAchievedMessageProvider.message(forGoal: goal)
+        if goal.date.differenceInDays(toDate: timeService.now) == 0 && goal.percentageCompleted < 1.0
```
Odrakir: Why do all these checks here if the message provider has a `timeService` instance and the `goal` as well?

teferi/UI/Modules/Goals/GoalList/GoalViewController.swift
```diff
@@ -72,10 +72,12 @@ class GoalViewController: UIViewController
             })
             .addDisposableTo(disposeBag)
         
-        viewModel.todaysGoal
+        viewModel.lastGoal
             .subscribe(onNext: { goal in
+                let messageAndCategoryVisibility = self.viewModel.messageAndCategoryVisibility(forGoal: goal)
```
Odrakir: Maybe instead of doing all this it's time to consider sending a reference to the `viewModel` into the header. Maybe constraining it to an interface, so it's access is limited to what it needs.

teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
sofokli1: i had it there at first but in the end it made sense for this logic to be in the viewModel (seamed too important to be elsewhere). will clean up the provider

teferi/UI/Modules/Goals/GoalList/GoalViewController.swift
sofokli1: just for passing 2 values doesn't feel like a good reason to break the pattern

teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
Odrakir: I know we are not strictly following any architecture, but in theory that kind of logic is not part of the VM.

teferi/UI/Modules/Goals/GoalList/Helper/GoalMessageProvider.swift
sofokli1: i thought i removed it 😱

teferi/UI/Modules/Goals/GoalList/GoalViewController.swift
Odrakir: 3 vars, the goal too.

teferi/UI/Modules/Goals/GoalList/GoalViewModel.swift
sofokli1: do you want me to move it back?

sofokli1: i also didn't want the `categoryVisible` to be decided in the provider

Odrakir: Everything should come from the provider, since there is one. Also, the category shouldn't be an independent `UIImageView`, it should be a `NSTextAttachment` so it's included in the text (in a `NSAttributedString`) and also so #887 doesn't happen. If you can't do that, at least include it somehow in the string and parse it in the view.

sofokli1: didn't have time for the `NSTextAttachment` approach

PR #894, opened 2017-11-21
Goal reminder notification

### LINKED ISSUE #834: Remind user to work on his/her goal
- Send notification 3 hours after the goal has been set if user set the goal before 8pm.
- When user set the goal after 8pm, don't send any notification after that.
- Show content based on which category user picked. Note: There'll be general notifications as well.

Family:
a) Spent some quality time with your Family already? Don’t forget your goal for today!
b) Always nice to squeeze some extra family time when possible. And also to check your goal status!

Fitness:
a) So what you’re doing today, cardio or some power training? Check how your Fitness goal is progressing.
b) You like to move it, move it? We like to track it, track it.
c) Time to move on. Or generally move a bit. Check your daily goal.

Food:
a) Mmmmmm, I think I’m up for some pizza. You? Remember to check your goal!

Leisure:
a) Finally, leisure time! What’s it gonna be – Netflix or games? Or both 😈?

Movement:
a) You like to move it, move it? We like to track it, track it.
b) Time to move on. Or generally move a bit. Check your daily goal.

Work:
a) Remember to do some work, work, work, work, work. And also check your goal, goal, goal, goal, goal!
b) Heigh-ho, heigh-ho, remember to do your work goal. See your progress.

Shopping:
a) Hide yo’ kids, hide yo’ credit cards. Just kidding, got some some shopping done today?

Odrakir:
Closes #834 

I'm not setting the texts as localizable strings because it would be an indirection mess at this point.


teferi/UI/Modules/Goals/NewGoal/NewGoalViewModel.swift
```diff
@@ -53,5 +56,37 @@ class NewGoalViewModel
...
+        }
+    }
+}
+
+fileprivate extension Category
+{
+    var notificationReminderText: String? {
+        switch self {
+        case .family:
+            return ["Spent some quality time with your Family today? Don’t forget your goal for today!",
```
sofokli1: all these text should be in `Constants.swift`

Odrakir: You mean in `Localizable.strings`, right? 

Why? What do we gain with that? Isn't it a bit of indirection overkill for an app that won't be getting multiple languages soon?

sofokli1: yes i meant `Localizable.strings`
First, it breaks consistency. Second, it is not bad to have all text in one place it makes for easy update of the strings (also for non devs). Third, i don't see a good reason to remove future proofing that is already in place. Are these good enough reasons?

Odrakir: I don't think a non-developer is ever going to touch that file. At least I've never seen anything like that in my whole career. And for us developers is much easier to see it the other way around, where we see the categories right next to the texts for each of them, rather than a bunch of weirdly named variables.

I'm mostly talking about this specific case, because each category has a different number of phrases and so the change would just imply replacing the text between the quotes with a property in that dynamically generated class.

The only time this makes sense, for me, is in the case of Localizing the app.

But yeah, ok, let's keep consistency (well, sort of, because there are still a bunch of hard coded texts)

PR #895, opened 2017-11-22
Settings tab design adjustments

### LINKED ISSUE #891: Issues in Settings
The Settings tab:
- Doesn't quite match the design (the header is bigger in design)
- Should (probably) also include the header shadow
- Is missing the build version at the bottom
- The feedback button in the nav bar should be removed for the whole app.

Odrakir:
Closes #891 

This PR removes the feedback button from navigation. Then moves the settings table view to an embeded table view controller so we can add the version number view at the bottom and still use static cells.


teferi/UI/Modules/Settings/Views/SettingsTableViewController.swift
```diff
@@ -0,0 +1,61 @@
+//
+//  SettingsTableViewController.swift
+//  teferi
+//
+//  Created by Ricardo Sánchez Sotres on 22/11/17.
+//  Copyright © 2017 Toggl. All rights reserved.
+//
```
amulware: :fire: please

PR #904, opened 2017-11-27
Running time edit

### LINKED ISSUE #901: Editing running TS doesn't save changes
1) Create new time slot manually (start time is 15:30 and duration is 0)
2) Tap manually created slot to open a detail view
3) Tap start time to edit:
img-4640

4) After editing, nothing happens (time slot was created at 15:30, and start time is still 15:30)
img-4641  img-4642

It happens only with currently running time slots.

Odrakir:
Closes #901 

I also added to this PR the updating of the running slot, as it's kind of related.

All this had to do with how I was handling the optional endTime. So now the persistence API had to change a little to adapt to that.


teferi/UI/Modules/Timeline/EditTimes/EditTimesViewModel.swift
```diff
@@ -40,49 +43,54 @@ class EditTimesViewModel: RxViewModel
             slot2 = slots.safeGetElement(at: slotIndex + 1)
         }
         
-        topSlot = Variable<TimeSlot?>(slot1)
-        bottomSlot = Variable<TimeSlot?>(slot2)
+        initialTopSlot = slot1!
```
sofokli1: this crashes when editing the start time of the first slot after installing the app

PR #906, opened 2017-11-28
:bug: Fix dismiss of detail VCs

### LINKED ISSUE #908: Bug when editing the start time of grouped slots.
The detail views are based on the start time to get the first slot for the grouped timeSlots (same for the single slot). when we edit the start time it is still trying to find the slot with the old time and it fails. In this situation it was suppose to dismiss the detail view. It was not dismissing back enough and thus ending in strange state with wrong data in the UI and a bit non responsive or in a crash (crash when editing the start time of the first grouped slot and trying to open that slot for a second time)

sofokli1:
closes #908


teferi/UI/Modules/Timeline/TimeSlotDetail/TimeslotDetailPresenter.swift
```diff
@@ -53,7 +53,25 @@ class TimeslotDetailPresenter: NSObject
     
     func dismiss()
     {
-        viewController.dismiss(animated: true)
+        var vcToDismiss: [UIViewController] = [viewController]
+        
+        while let vc = vcToDismiss.last?.presentedViewController {
```
Odrakir: You are modifying an array while iterating over it, this is not a good idea generally. Find another way to do it.

sofokli1: i agree with the feeling about modifying an array while iterating it but in this situation it is not doing that. since it only looks at the last item in the array it is not iterating the array it is like having a var that is changed for every iteration of the while loop.

sofokli1:
regarding the 10 seconds refresh it is most likely related to https://github.com/toggl/superday/blob/fix-bug-when-changing-start-time/teferi/UI/Modules/Timeline/TimelineViewModel.swift#L56
to me it makes more sense to think of something better for the refresh than to change the logic in the details view

teferi/UI/Modules/Timeline/TimeSlotDetail/TimeslotDetailPresenter.swift
Odrakir: I know it works in this situation. But it's not a good idea, it's easy to break and easy to get you into an infinite loop. We should stay away from this kind of code. But anyway, I think this whole PR needs changing.

Odrakir:
Sure it's because of that. We are sending an Observable used in one viewModel to another viewModel which has other needs. I think the refresh logic works great for the timeline, but for the detail view we need something else. In the end I think it's not a good idea to share that Observable, I don't think it's much what we save in terms of performance, and in any case is premature optimization.

PR #932, opened 2017-12-06
Fix start time editing

### LINKED ISSUE #908: Bug when editing the start time of grouped slots.

sofokli1:
Closes #908 

This approach propagates the start time from the VC that the change happens down to the details view for grouped and single slot.

An other approach for this kind of issues would be to have identifiers per timeSlot. It seemed too big of a change for this PR but something worth thinking about.


teferi/Extensions/ArrayExtensions.swift
```diff
@@ -109,6 +109,35 @@ extension Array where Element : Hashable
     }
 }
 
+extension Array where Element == TimeSlot
```
Odrakir: This extension is too specific to be part of Array Extensions (at least of this file). The same happens with the next one, but let's leave it like this for now.

teferi/UI/Modules/Timeline/TimeSlotDetail/TimeslotDetailViewModel.swift
```diff
@@ -7,40 +7,70 @@ class TimeslotDetailViewModel
...
         self.metricsService = metricsService
         self.smartGuessService = smartGuessService
         self.timeService = timeService
         self.categoryProvider = DefaultCategoryProvider(timeSlotService: timeSlotService)
         
-        timelineItemsObservable
-            .map(filterSelectedElement(for: startDate))
+        updateStartDateSubject.asObservable().subscribe(onNext: { [unowned self] (date) in
+            self.startDate = date
+            print(self.startDate, #function)
```
Odrakir: print

teferi/UI/Modules/Timeline/TimeSlotDetail/TimeslotDetailViewModel.swift
```diff
@@ -78,41 +128,40 @@ class TimeslotDetailViewModel
...
+            return TimelineItem(withTimeSlots: [timeSlotToShow],
+                                category: timeSlotToShow.category,
+                                duration: timeSlotToShow.duration != nil ?
+                                    timeSlotToShow.duration! :
+                                    self.timeService.now.timeIntervalSince(timeSlotToShow.startTime),
+                                isRunning: isRunning)
+        }
+        else
+        {
+            print(startDate, #function)
```
Odrakir: print

PR #935, opened 2017-12-07
:bug: Fix retain cycle in details view

### LINKED ISSUE #931: Crash due to memory issue
After editing time slots for some time the app crashes and XCode just says "terminating due to memory issue"

sofokli1:
Fixes retain cycle issue


sofokli1:
do you still have this problem with the changes from #932?

sofokli1:
This PR now fixes only the retain cycle issue

PR #936, opened 2017-12-08
Add firebase

### LINKED ISSUE #917: Replace Fabric Answers with Firebase for analytics
Firebase brings more features for analyzing that info, so we need to move to it.

Instructions: https://firebase.google.com/docs/analytics/ios/start

sofokli1:
Closes #917


teferi/Services/Implementations/Monitoring/FirebaseMetricsService.swift
```diff
@@ -0,0 +1,30 @@
...
+            guard let fileopts = FirebaseOptions.init(contentsOfFile: filePath!)
+                else { assert(false, "Couldn't load config file"); return }
+            FirebaseApp.configure(options: fileopts)
+        #endif
+    }
+    
+    ///Used to send custom events to the framework
+    func log(event: CustomEvent)
+    {
+        #if ADHOC || APPSTORE
```
Odrakir: `#if !DEBUG`?

teferi/Services/Implementations/Monitoring/FirebaseMetricsService.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import Firebase
+
+class FirebaseMetricsService : MetricsService
+{
+    ///Perform any framework specific initialization
+    func initialize()
+    {
+        #if ADHOC
```
Odrakir: Why do we want this for ad_hoc?

sofokli1: to separate adhoc from appstore so that we can see what happens in testing and in live separately. we had it in fabric as well.

teferi/Services/Implementations/Monitoring/FirebaseMetricsService.swift
sofokli1: i made it like that for readability and making it clear when it is used. do you want it changed?

Odrakir: I think it's more readable the other way around, yes.

teferi/Services/Implementations/Monitoring/FirebaseMetricsService.swift
Odrakir: We weren't doing anything specifically with adhoc previously, we just weren't sending any events for debug. But it's ok, leave it.

PR #939, opened 2017-12-11
Log finished goals

### LINKED ISSUE #921: Track of how many users achieve or fail the goal
We want to know of how many users who set a goal, manage to achieve or fail it.

sofokli1:
closes #921


teferi/AppDelegate.swift
```diff
@@ -201,6 +203,32 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         
         return notification.date
     }
+    
+    private func logFinishedGoals()
```
Odrakir: Please, move this out the app delegate and inside `DefaultGoalService`(injecting the appropiate services)

PR #951, opened 2017-12-18
🚧📐Make tab bar translucent

### LINKED ISSUE #944: Make tab bar translucent
Currently, tab bar at the bottom isn't translucent though it should be.

sofokli1:
closes #944


teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -78,7 +78,7 @@ class TimelineViewController : UIViewController
         tableView.showsHorizontalScrollIndicator = false
         tableView.register(UINib.init(nibName: "TimelineCell", bundle: Bundle.main), forCellReuseIdentifier: TimelineCell.cellIdentifier)
         tableView.register(UINib.init(nibName: "ShortTimelineCell", bundle: Bundle.main), forCellReuseIdentifier: ShortTimelineCell.cellIdentifier)
-        tableView.contentInset = UIEdgeInsets(top: 34, left: 0, bottom: 49, right: 0)
+        tableView.contentInset = UIEdgeInsets(top: 34, left: 0, bottom: 49 + (tabBarController?.tabBar.frame.height ?? 0), right: 0)
```
Odrakir: Same here this should be using margins/layout guides/safe areas. Not sure if this could be problematic in the iPhone X...

teferi/UI/Common/Views/Alert.swift
```diff
@@ -33,8 +33,10 @@ class Alert: UIView
         containerView.addSubview(shadow)
         containerView.addSubview(effectBackground)
         
+        print(bottomBarHeight)
         containerView.snp.makeConstraints { make in
-            make.leading.trailing.bottom.equalToSuperview().inset(16)
```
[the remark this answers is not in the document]
sofokli1: you mean to do something like `make.bottom.equalTo(self.bottomLayoutGuide.snp.top)` or `make.bottom.equalTo(self.safeAreaLayoutGuide.snp.bottom)`? the problem is that `Alert` is a view and superview is also a view. correct me if i am wrong,`self.safeAreaLayoutGuide` and `self.bottomLayoutGuide` are properties of `ViewControllers`, so i can't get a hold of them from the alert. Do you have any suggestion for this?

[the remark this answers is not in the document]
sofokli1: never mind, i found an other way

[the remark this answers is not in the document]
Odrakir: I think it's ok to change the API of `Alert` to receive a `UIViewController` instead of a `UIView`.

sofokli1:
@Odrakir How are you testing this? i am testing on iphone x simulator and other simulator and looks translucent. Can you try cleaning the project please?

Odrakir:
I'm sorry. Wrong branch.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -78,7 +78,7 @@ class TimelineViewController : UIViewController
         tableView.showsHorizontalScrollIndicator = false
         tableView.register(UINib.init(nibName: "TimelineCell", bundle: Bundle.main), forCellReuseIdentifier: TimelineCell.cellIdentifier)
         tableView.register(UINib.init(nibName: "ShortTimelineCell", bundle: Bundle.main), forCellReuseIdentifier: ShortTimelineCell.cellIdentifier)
-        tableView.contentInset = UIEdgeInsets(top: 34, left: 0, bottom: 49, right: 0)
+        tableView.contentInset = UIEdgeInsets(top: 34, left: 0, bottom: 49 + self.view.layoutMargins.bottom, right: 0)
```
Odrakir: If you are setting the `contentInset` in `viewDidLayoutSubviews` you don't need to do it here as well.

sofokli1: it may look like it but if i only do it in `viewDidLayoutSubviews` the first slot will be under the activity bar.

Odrakir: I see, but this is very error prone, in the future we'll probably change one but not the other.

But anyway, this still doesn't work right. The bottom margin is shorter than what it used to and also in some cases it cuts the timeline..

This one is scrolled to the bottom: 👇 
img_f93f5dbdcae9-1

sofokli1: `self.view.layoutMargins.bottom` gives `80.0` when on iphone x and `0.0` when on other iphone simulators. Unless you have any suggestion for this, i recommend going back to using `tabBarController.tabBar.frame.height` for the `tableView.contentInset` and remove the other call in `viewDidLayoutSubviews`.

Odrakir:
Ok, we are going to use the tab bar's frame for now. But this still doesn't quite work. I've been doing some research and it is due to the container view probably (and maybe also due to the pager view). Those two prevent the insets to be forwarded from the main view. I think the best thing to do here is to replace the whole timeline by a collection view with table view cells. I'll create an issue.

PR #953, opened 2017-12-20
🚧Bring back old collapse logic

### LINKED ISSUE #947: Allow expanding/collapsing time slots again
- [ ] Tapping a slot that contains more than one slot, expands this slot.
- [ ] Tapping the arrow at the bottom, collapses the slot again.
- [ ] Each expanded slot is tappable and opens edit view of this specific slot. _This means edit view is only for one slot only, not for multiple slots anymore._
- [ ] Both, regular and movement slots are expandable/collapsible.
- [ ] Do not show start time or total duration next to the main slot when slots are expanded. Start time and duration will be displayed next to each expanded slot.

### Expanded regular slot
timeline - expanded regular slot 1x

Zeplin: https://zpl.io/VQMpzN5

### Expanded movement slot
timeline - expanded movement slot 1x

Zeplin: https://zpl.io/bAO6PRp

sofokli1:
closes #947 

To handle the new UI for the expansion of the grouped slots I added 3 new tableview cells 1 for the slots when expanded, 1 for the to be used as header for the grouped slots (showing the category) and 1 for the collapse button. I removed the expand and collapse buttons from the old timeline cell since the logic is handled in `tableView.rx.modelSelected`. The new category only cell and the collapse cell are timelineItems with specific flags so that `constructCell` knows what to display. Also the detail view logic for multiple slots handling has been removed.

I leave open for discussion the possibility to use an enumerator instead of flags to describe the different cell types inside the `timelineItem`


teferi/UI/Modules/Timeline/Views/ExapndedTimelineCell.swift
```diff
@@ -0,0 +1,59 @@
+import UIKit
+
+class ExapndedTimelineCell: UITableViewCell
```
Odrakir: Typo in class name.

teferi.xcodeproj/project.pbxproj
```diff
@@ -234,6 +234,12 @@
 		AC3901F41E5665CF008063C7 /* CategoryButtonsHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC3901F31E5665CF008063C7 /* CategoryButtonsHandler.swift */; };
 		AC449C311E92BC8C00847957 /* PermissionViewModelProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC449C301E92BC8C00847957 /* PermissionViewModelProtocol.swift */; };
 		AC449C3B1E95690700847957 /* LocationPermissionViewModelTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC449C3A1E95690700847957 /* LocationPermissionViewModelTests.swift */; };
+		AC4667661FFE485B00C3BD5E /* ExapndedTimelineCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC4667641FFE485B00C3BD5E /* ExapndedTimelineCell.swift */; };
+		AC4667671FFE485B00C3BD5E /* ExapndedTimelineCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = AC4667651FFE485B00C3BD5E /* ExapndedTimelineCell.xib */; };
```
Odrakir: Typo in filename.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -170,6 +171,31 @@ class TimelineViewController : UIViewController
...
+    {
+        if item.isSubSlot
+        {
+            presenter.showEditTimeSlot(with: item.startTime)
+        }
+        else if item.isSubSlotCollapse
+        {
+            viewModel.collapseSlots()
+        }
+        else if item.isSubSlotHeader
```
Odrakir: What does this if do?

Also, I thought we agreed (though not enforced yet) on only use new line for braces in functions and classes definitions.

teferi/UI/Modules/Timeline/TimelineViewModel.swift
```diff
@@ -141,21 +156,67 @@ class TimelineViewModel
     
     private func toTimelineItems(fromTimeSlots timeSlots: [TimeSlot]) -> [TimelineItem]
     {
-        return timeSlots.toTimelineItems(timeSlotService: timeSlotService, isCurrentDay: isCurrentDay)
+//        return timeSlots.toTimelineItems(timeSlotService: timeSlotService, isCurrentDay: isCurrentDay)
```
Odrakir: Comment.

teferi/UI/Modules/Timeline/TimelineViewController.swift
sofokli1: - This if is used for filtering (meaning not doing anything). I could have put a guard but it made more sense like this so that the ifs cover clearly all possibilities.

- I can't recall a clear agreement on this. I am just used to do it this way, i find it more clear. I am not expecting everyone to do it like this. Also the guideline still describe it like this.

Odrakir:
But you didn’t do anything with the comment about the TimelineItems. Did you? The current solution with flags is not the best way to do it.

sofokli1:
I didn't do anything about that because it seemed like too much of a refactoring for the purpose of this PR. In general I like the idea of having a protocol per tableview/collectionview that the models implement. In this situation I fail to see how this would improve the situation of the `if` statements of the flags. What would we check for in the protocol?

Odrakir:
I think the current solution with flags inside `TimelineItem` is not optimal. My idea is to turn `TimelineItem` into a protocol and then instead of checking for all those flags, just check for specific types in a switch statement. Maybe there's another way of doing this with enums, as you mentioned (what was that idea?), anyway in my opinion this needs to be improved.

sofokli1:
The less invasive idea of enum was to use a `TimelineItemType` with specific cases of `normal`, `groupedHeader` and so on to replace the flags. With this specific types, when checking with a `switch` statement there is no way a case can be missed and makes for a more clear approach. Easy to spot forgot places at compile time when new types are added. It will reduce the flags to just 1 property and replace the error prone `if`s to delightful `switch`es 😌

Odrakir:
Oh, but you still have the properties `timeSlots`, `duration`, `category`and so on for all cell models, even when they don't need or understand them. Better go with the protocol approach.

sofokli1:
So what properties should the protocol require? Or maybe none at all, just so that we can fit different models structs in one array?

Odrakir:
Try to code it and see which ones are common... I'm guessing `category`, but I might be wrong.

teferi/Models/SectionItem.swift
```diff
@@ -0,0 +1,53 @@
...
+{
+    case timelineItemType(item: TimelineItem)
+    case commuteTimelineItemType(item: TimelineItem)
+    case expandedCommuteHeaderTimelineItemType(item: TimelineItem, showStartAndDuration: Bool)
+    case expandedHeaderType(item: TimelineItem)
+    case expandedTimelineItemType(slot: TimeSlot, duration: TimeInterval, hasSeparator: Bool)
+    case expandedCollapseFooterType(color: UIColor)
+    
+    
+    var identity: String
```
Odrakir: Identity and equality are two different things. Identity identifies each and every item and differentiates it from any other item. Equality tells you when an item has changed even if it has the same identity. That's why we used startTime as the identity previously.

Now you should still have startTime as identity, followed by "title" or "button" to differentiate those two cells which will have the same startTime as the first item. And then check for the properties that imply visual changes for equality (category, isRunning, elapsedTimeText and wether is expanded or not).

This will also fix the fading animation when expanding/contracting as well.

teferi/Models/SectionItem.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    static func ==(lhs: SectionItem, rhs: SectionItem) -> Bool
+    {
+        return lhs.identity == rhs.identity
+    }
+}
+
+extension SectionItem
+{
+    var timelineItem: TimelineItem?
```
Odrakir: I don't see the point in this. Correct me if I'm wrong, but you are only using it for `TimelineViewModel`line 106, where it makes more sense to check for the specific enum cases that can be expanded, as not all of them with this property set to not nil can be expanded.

teferi/Models/TimeSlot.swift
```diff
@@ -10,6 +10,11 @@ struct TimeSlot
     let categoryWasSetByUser: Bool
     let categoryWasSmartGuessed: Bool
     let activity: MotionEventType?
+    
+    var identifier: String
```
Odrakir: I mentioned this in another comment, but identity shouldn't be produced like this. Also, in any case this is something you need just for the tableView representation of this data, that info shouldn't get this far.

teferi/Models/TimelineItem.swift
```diff
@@ -23,11 +23,21 @@ struct TimelineItem
     {
         return timeSlots.count > 1
     }
+    
+    var identifier: String
```
Odrakir: I mentioned this in another comment, but identity shouldn't be produced like this.

teferi/Models/TimelineItem.swift
```diff
@@ -48,6 +58,15 @@ extension TimelineItem
             duration: self.duration,
             shouldDisplayCategoryName: self.shouldDisplayCategoryName,
             isLastInPastDay: !isCurrentDay,
-            isRunning: isCurrentDay)
+            isRunning: isCurrentDay
+        )
+    }
+    
+    static func with(timeSlot: TimeSlot, timeSlotService: TimeSlotService) -> TimelineItem
```
Odrakir: You are setting `shouldDisplayCategoryName` to true, but nowhere in the signature of the method is there any hint of that. I would remove this method and create the `TimelineItem` directly.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -184,6 +185,37 @@ class TimelineViewController : UIViewController
             .addDisposableTo(disposeBag)
     }
     
+    private func handleTableViewSelection(sectionItem: SectionItem)
+    {
+        switch sectionItem {
+        case .timelineItemType(let item),
+             .commuteTimelineItemType(let item):
```
Odrakir: Commute slots are not expanding for me 🤷‍♂️

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -184,6 +185,37 @@ class TimelineViewController : UIViewController
...
+            if item.timeSlots.count > 1
+            {
+                viewModel.expandSlots(sectionItem: sectionItem)
+            }
+            else
+            {
+                presenter.showEditTimeSlot(with: item.startTime)
+            }
+            
+        case .expandedCommuteHeaderTimelineItemType(_, _),
```
Odrakir: I'm ok with not using default to not miss any cases, but put empty cases at the end.

teferi/UI/Modules/Timeline/Views/CategoryCell.swift
```diff
@@ -0,0 +1,35 @@
...
+    private(set) var timelineItem: TimelineItem?
+    
+    var editClickObservable : Observable<TimelineItem> {
+        return self.categoryButton.rx.tap
+            .mapTo(self.timelineItem)
+            .filterNil()
+            .asObservable()
+    }
+    
+    private(set) var disposeBag = DisposeBag()
```
Odrakir: I'm not 100% sure about this one, but shouldn't the disposeBag be created inside `configure` so it disposes the previous subscription when reusing the cell?

teferi/UI/Modules/Timeline/Views/CollapseCell.swift
```diff
@@ -0,0 +1,16 @@
...
+class CollapseCell: UITableViewCell
+{
+    static let cellIdentifier = "CollapseCell"
+    
+    @IBOutlet private weak var collapseImageView: UIImageView!
+    
+    func configure(color: UIColor)
+    {
+        let image = UIImage(asset: Asset.icCollapse).withRenderingMode(.alwaysTemplate)
+        collapseImageView.image = image
```
Odrakir: You are setting the image in storyboard as well, so it's better to just:
`collapseImageView.image = collapseImageView.image!.withRenderingMode(.alwaysTemplate)`
and also you just need to do it once in `awakeFromNib`, for example.

Odrakir:
Also, just from a UX point of view... Can we make the hit area of the collapse button a little bit bigger?

sofokli1:
the cell will have to be bigger

Odrakir:
Judging by the storyboard margins can be moved a little bit, without changing the design.

teferi/Models/SectionItem.swift
sofokli1: ok, but i also need it for tests

Odrakir: We are not really worrying about tests right now, but even so, we shouldn't include code just for them. If you really need to have those tests, unwrap it there.

teferi/Models/SectionItem.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+import RxDataSources
+
+enum SectionItem : IdentifiableType, Equatable
```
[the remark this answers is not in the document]
sofokli1: how about `SectionItem` -> `TimelineSectionItem`?

teferi/Models/SectionItem.swift
sofokli1: I was just trying to avoid boilerplate code

Odrakir: But this thing you did can cause problems when updating the table view. Did I clarify why?

teferi/Models/TimelineItem.swift
```diff
@@ -23,11 +23,21 @@ struct TimelineItem
...
+    var identifier: String
+    {
+        return startTime.description + category.rawValue + "\(isRunning)" + elapsedTimeText + "\(containsMultiple)"
+    }
 }
 
 extension TimelineItem
 {
-    init(withTimeSlots timeSlots: [TimeSlot], category: Category, duration: TimeInterval, shouldDisplayCategoryName: Bool = false, isLastInPastDay: Bool = false, isRunning: Bool = false)
+    init(withTimeSlots timeSlots: [TimeSlot],
```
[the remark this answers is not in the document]
sofokli1: it is used in the extension below

teferi/Models/SectionItem.swift
Odrakir: You are right, but otherwise we should include an ID for every item... Maybe we can do that in the future. For now this is good enough, as the timeline is generated again when a change in the startTime happens.

teferi/Models/TimelineItem.swift
```diff
@@ -1,53 +1,78 @@
...
-    let category: Category
-    let duration : TimeInterval
-    let shouldDisplayCategoryName : Bool
-    let isLastInPastDay : Bool
-    let isRunning: Bool
+    case slotTimelineItemType(item: SlotTimelineItem)
+    case commuteSlotTimelineItemType(item: SlotTimelineItem)
+    case expandedCommuteHeaderSlotTimelineItemType(item: SlotTimelineItem)
+    case expandedTitleType(item: SlotTimelineItem)
+    case expandedSlotTimelineItemType(slot: TimeSlot, duration: TimeInterval, hasSeparator: Bool)
```
Odrakir: Instead of `TimeSlot` and `duration` use a `SlotTimelineItem` here as well. This is what it's for, to abstract away `TimeSlot`, event if it's just for one of them.

teferi/Models/TimelineItem.swift
```diff
@@ -1,53 +1,78 @@
...
     
-    var endTime: Date?
+    var identity: String
     {
-        return timeSlots.last!.endTime
+        switch self {
+        case .slotTimelineItemType(let item):
+            return item.startTime.description
+        case .commuteSlotTimelineItemType(let item):
+            return item.startTime.description + "commute"
```
Odrakir: I don't think this will cause any problem, but still "commute" is not needed here as it won't be sharing startTime with any other item. We only need to differentiate the titles and the collapse button, which don't have its own startTime.

teferi/Models/TimelineItem.swift
```diff
@@ -1,53 +1,78 @@
...
+        case .expandedCommuteHeaderSlotTimelineItemType(let item):
+            return item.startTime.description + "commuteTitle"
+        case .expandedTitleType(let item):
+            return item.startTime.description + "expandedTitle"
+        case .expandedSlotTimelineItemType(let slot, _, _):
+            return slot.startTime.description + "expanded"
```
Odrakir: Same here, "expanded" is not needed.

teferi/Models/TimelineItem.swift
```diff
@@ -1,53 +1,78 @@
...
+        case .expandedCollapseType(let color):
+            return color.hexString + "expandedCollapse"
+        }
     }
     
-    var containsMultiple: Bool
+    static func == (lhs: TimelineItem, rhs: TimelineItem) -> Bool
```
Odrakir: The intention of this should be to compare `TimelineItem` instances in order to see if any visual changes need to be done. With our current implementation that depends on the specific enum the changes are ocurring. The equatable extensions on `TimelineItem` and `TimeSlot` are confusing because that's not what you are trying to compare.

After you change `expandedSlotTimelineItemType` to take a `SlotTimelineItem` this is the code you should use:

```swift
    
    static func == (lhs: TimelineItem, rhs: TimelineItem) -> Bool
    {
        switch (lhs, rhs) {
        case (.slotTimelineItemType(let lhsItem), .slotTimelineItemType(let rhsItem)),
             (.expandedSlotTimelineItemType(let lhsItem, _), .expandedSlotTimelineItemType(let rhsItem, _)),
             (.commuteSlotTimelineItemType(let lhsItem), .commuteSlotTimelineItemType(let rhsItem)):
            return lhsItem.duration == rhsItem.duration
                && lhsItem.isRunning == rhsItem.isRunning
                && lhsItem.elapsedTimeText == rhsItem.elapsedTimeText
                && lhsItem.category == rhsItem.category
                && lhsItem.containsMultiple == rhsItem.containsMultiple
        case (.expandedTitleType(let lhsItem), .expandedTitleType(let rhsItem)),
             (.expandedCommuteHeaderSlotTimelineItemType(let lhsItem), .expandedCommuteHeaderSlotTimelineItemType(let rhsItem)):
            return lhsItem.category == rhsItem.category
                && lhsItem.duration == rhsItem.duration
        case (.expandedCollapseType(let lhsColor), .expandedCollapseType(let rhsColor)):
            return lhsColor == rhsColor
        default:
            return false
        }
    }
```

teferi/UI/Modules/Timeline/Views/CollapseCell.xib
```diff
@@ -12,15 +12,15 @@
     <objects>
         <placeholder placeholderIdentifier="IBFilesOwner" id="-1" userLabel="File's Owner"/>
         <placeholder placeholderIdentifier="IBFirstResponder" id="-2" customClass="UIResponder"/>
-        <tableViewCell contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" rowHeight="20" id="KGk-i7-Jjw" customClass="CollapseCell" customModule="teferi" customModuleProvider="target">
-            <rect key="frame" x="0.0" y="0.0" width="320" height="20"/>
+        <tableViewCell contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" rowHeight="22" id="KGk-i7-Jjw" customClass="CollapseCell" customModule="teferi" customModuleProvider="target">
```
Odrakir: 22 points is not enough still, it's hard to hit.

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -224,29 +253,75 @@ class TimelineViewController : UIViewController
...
+            
+        case .commuteSlotTimelineItemType(let item):
+            
             let cell = tableView.dequeueReusableCell(withIdentifier: ShortTimelineCell.cellIdentifier, for: indexPath) as! ShortTimelineCell
-            cell.timelineItem = item
+            cell.configure(slotTimelineItem: item)
+            cell.selectionStyle = .none
+            return cell
+            
+        case .expandedCommuteTitleSlotTimelineItemType(let item):
```
Odrakir: This one is exactly the same as the previous one, can't we merge them?

teferi/UI/Modules/Timeline/TimelineViewController.swift
```diff
@@ -224,29 +253,75 @@ class TimelineViewController : UIViewController
         tableView.scrollToRow(at: scrollIndexPath, at: .bottom, animated: true)
     }
     
-    private func constructCell(dataSource: TableViewSectionedDataSource<TimelineSection>, tableView: UITableView, indexPath: IndexPath, item:TimelineItem) -> UITableViewCell
+    private func constructCell(dataSource: TableViewSectionedDataSource<TimelineSection>, tableView: UITableView, indexPath: IndexPath, sectionItem: TimelineItem) -> UITableViewCell
     {
-        if item.category == .commute {
-         
+        switch sectionItem {
```
Odrakir: This should be `timelineItem`

teferi/Models/TimelineItem.swift
```diff
@@ -1,53 +1,53 @@
 import Foundation
+import RxDataSources
 
-struct TimelineItem
+enum TimelineItem : IdentifiableType, Equatable
```
Odrakir: Name of cases for this enum are getting a bit out of hand, I think this is an easy change, how about calling them:

```swift
    case slot(item: SlotTimelineItem)
    case commuteSlot(item: SlotTimelineItem)
    case expandedCommuteTitle(item: SlotTimelineItem)
    case expandedTitle(item: SlotTimelineItem)
    case expandedSlot(item: SlotTimelineItem, hasSeparator: Bool)
    case collapseButton(color: UIColor)
```

Odrakir: Correctly ordered:

```swift
    case slot(item: SlotTimelineItem)
    case commuteSlot(item: SlotTimelineItem)
    case expandedTitle(item: SlotTimelineItem)
    case expandedCommuteTitle(item: SlotTimelineItem)
    case expandedSlot(item: SlotTimelineItem, hasSeparator: Bool)
    case collapseButton(color: UIColor)
```

teferi/UI/Modules/Timeline/Views/CollapseCell.swift
```diff
@@ -0,0 +1,28 @@
+import UIKit
+import RxSwift
+
+class CollapseCell: UITableViewCell
+{
+    static let cellIdentifier = "CollapseCell"
+    
+    @IBOutlet private weak var collapseImageView: UIImageView!
+    @IBOutlet private weak var collapseButton : UIButton!
```
Odrakir: The button doesn't increase the hit area of the cell, in any case it decreases it. The only way I can think of to improve this is to change the upper margin to make this cell higher and in order to keep the design you'll then need to change the expanded items as well.

sofokli1: The button was not added to increase the hit area as described in the comment https://github.com/toggl/superday/pull/953#issuecomment-359519652

Odrakir: I see. I misunderstood. Anyway, that wasn't really the problem. The problem is it's still too hard to hit the collapse arrow. I suggest merging that cell with the last slot so it's easy to increase the hit area.

teferi/UI/Modules/Timeline/TimelineViewController.swift
sofokli1: they are not the same

Odrakir: Oh, ok.

PR #956, opened 2017-12-22
Adds doc about the smart guessing system

Odrakir:
Just as a reference for us and maybe support as well.


docs/SmartGuessing.md
```diff
@@ -0,0 +1,137 @@
...
+
+The current implementation of the Location Service needs to be streamlined to send only the most accurate location for every significant location change.
+
+For every location event we need to store **latitude, longitude and timestamp**. Storing the rest of the information like altitude, accuracy, speed... it's probably a good idea, but we are not making a use of it right now.
+
+### Annotated Motion Events
+
+When the user opens the app we get all CoreMotion activity events since the last time he opened it (a week tops). As this data has very heigh granularity we first make a first parsing pass to get rid of unknown motion activity events.
+
+At this point the work flow splits in two paths. First we get those motion events and merge them as much as we can to store them in the DB. Then with those same un-merged events we merge them with the location events in that timeframe. Which leaves us with a series of motion activity events each of them with a timestamp and a location.
```
sofokli1: Since you are using the un-merged motion events to create the timeSlots, where do you use the merged motion events that you store in the database? And for what?

Odrakir: The idea is to have the following tables in the DB: Motion, Location and Timeslots. The first two are self explanatory and the last one only stores dates and categories (and also wether it's been edited by the user or not). So then from the app we can merge all that info into a Timeslot which has a location (or locations) and the different motion activities in its time frame.

Then if the user ever edits start/end times we can just check those tables again to see motion and location events in the new time frame.

Odrakir: I store those events merged because at that point I don't need the un merged version of them anymore.

For example, if I have a 6 hour still motion event, followed by a 20 minute one, the parsing algorithm created two separated slots (the first one is probably sleep). But once that's done I can persist them as a merged big slot as I don't need that information anymore (it's stored in TimeSlots)

sofokli1: Just to make sure that i understand it correctly. You mean to have the TimeSlots table (that has dates and categories) reference the other 2 tables for location and motion, right?

Odrakir: No. The three tables are independent and only merged when we fetch time slots.

docs/SmartGuessing.md
```diff
@@ -0,0 +1,137 @@
...
+
+After the previous step we end up with a timeline divided in time slots in which we believe the user was doing different specific activities, but we don't know which activities yet. Although we could make some guesses looking at the motion events (if it's auto we guess commute, for example), it's probably a better solution to let the machine learning algorithm do most or all of the work (a commute by car may be categorize as work for the user, for example)
+
+To guess the category for every time slot we will use a machine learning algorithm called **Support Vector Machine**. We first train the algorithm with a list of previously categorized events and then the algorithm can guess the category of new events.
+
+For each slot we feed to the algorithm: the start time, the duration, the day of the week, the latitude and longitude and the top motion activity event in that timeframe.
+
+For the training phase we use first the _standard day_ we hard coded into the app and then whichever time slots the user categorized hims self or voted as correct using the rating system in the app. Depending on the amount of data we might consider using also non voted and non rated slots.
```
amulware: I'd be careful with this. If we don't take un-voted data, we might be missing out on a lot if the user cannot be bothered. Especially if it's not communicated to the user that they should move, they might never bother. Though if it is communicated, they might be bothered in turn. My intuition says to use all recent data, whether it was guessed or not, but maybe there is a way to have user entered/corrected/upvoted data count for more, to give more engaged users better results.

Odrakir: My idea is to feed the algorithm the slots we know the user is happy with. How do we know which ones are those? We take those in which the user edited the categories and also those in which he edited the start or end times. Plus those in days he rated as "ok".

I think that, together with the standard day should be ok.

The generated timeline is either ok, so we can keep using those slots as a guide, or it's not ok, so the user should correct it and we have new data to work with. If the user doesn't care enough to correct our guesses... Maybe he just shouldn't use this app?

amulware: Your last paragraph makes some assumptions and asks some good questions. We may wanna discuss what exactly our approach here is. i.e. whether we expect users to actively engage with the app a lot or not.

PR #963, opened 2018-01-09
Add new persistency layer

### LINKED ISSUE #962: Add new persistency layer
The basic idea is to end up having a layered architecture. We are going to do this one step at a time, so a big refactor is not needed (or rather, can be done gradually).

In the end we'll have different model types for the three main layers:
- Persistence
- Business logic
- Presentation

Until now there was a more or less 1 to 1 relationship between the models of those layers (not always, though). Meaning the TimeSlot we stored in the DB is the same one we use for the logic and (more or less) the same one we show in the UI. 

But this is changing with the new smart guessing system. The objects we'll store are different than the ones we use for the business logic and those are also different from the ones used in presentation. That's when we need the layered approach.

For example, the DB will store:
- TimeSlotMOs (timestamp, category and edited flag)
- LocationEvents
- MotionEvents

Which on fetch, will be mapped to the business logic model:
- TimeSlot(containing end times and the annotated motion events in its timeframe)

And these will be transformed for display in the timeline to:
- TimelineItem(which might or might not contain multiple TimeSlots)

Other than a public API, display doesn't need to know anything about business logic and business logic doesn't need to know anything about the persistence. 

This first issue takes care of the persistence layer, creating the models and the necessary classes to store those objects.

Odrakir:
Closes #962 

So, the basic idea is DB models are fetched through an entity called `CoreDataResource`, this type of entity describes the fetch request and the parsing of the `NSManagedObject` into the specific model we are trying to obtain.

We have an abstraction of the CoreData data base that implements a fetch method which takes a resource and returns an observable of whatever model the resource is for.

The "PM" (persistence model) structs are only for persistency and other than methods to decode and encode from/to `NSManagedObject`, they also have static methods to build the resources needed for specific queries.

That way the fetching of models can be done as:
```swift
persistency.fetch(TimeSlotPM.all(fromDate: lastWeek, toDate: today))
```

which returns an observable.

For now I only implemented methods to fetch entities between two dates, more methods will be done when needed.


teferi/Services/Implementations/New Persistency/CoreDataPersistency.swift
```diff
@@ -0,0 +1,41 @@
...
+
+enum PersistencyError: Error
+{
+    case noResults
+    case couldntParse
+}
+
+class CoreDataPersistency
+{
+    let managedObjectContext: NSManagedObjectContext
```
sofokli1: Shouldn't this be private?

Odrakir: Yep, fixed. Please review.

teferi/Models/Persistency/TimeSlotPM.swift
```diff
@@ -0,0 +1,42 @@
...
+extension TimeSlotPM: PersistencyModel
+{
+    static var entityName: String = "TimeSlot"
+    
+    init(managedObject: NSManagedObject) throws
+    {
+        guard let startTime = managedObject.value(forKey: "startTime") as? Date,
+            let categoryString = managedObject.value(forKey: "category") as? String,
+            let category = Category(rawValue: categoryString) else {
+                throw PersistencyError.couldntParse
```
sofokli1: This comment goes for the other similar files as well.
If the values that we are trying to get are nil that means that we didn't save any value for that key (or did not set any default value in the coredata scheme) so I do not think we should handle it as a `couldntParse` error since it parses just fine if the value is nil. If `managedObject.value(forKey: "...")` throws than it makes sense to throw a `couldntParse` error because in that case something has gone completely wrong and the `key` is not part of the `managedObject`.

Wouldn't it make sense to have an other error for the nil values (if you think it is needed) as well as the `couldntParse` for when the `managedObject.value(forKey` throws?

Odrakir: This kind of check is there for mandatory properties only, meaning properties that can't be nil, so if they are nil then there's a problem parsing those results. In the case of "editedByUser" we do allow it to be nil.

I think the error is correct.

sofokli1: I get that some keys should always have a value. What I am trying to saying is that if the values is nil it is not a parsing problem, it is a saving problem. The error naming is misleading. Something like `missingManadatoryValue`makes more sense. Your call

Odrakir: We don't need more granular errors for now, let's leave it like it is.

PR #964, opened 2018-01-10
🚧📐Fix calendar UI glitch

### LINKED ISSUE #933: Calendar days don't show correct colors
It happens often when moving between months.

As you can see the first screenshot is missing all days, the second (after moving to November and back) still misses Monday and Tuesday.

img_4088

sofokli1:
Closes #933

I think calculations for the activity view are off because the frames of the views are not known yet when reload data is called. The solution i tried was to reload the data after the animations run and all the frames are set


sofokli1:
Please provide screenshot of the calendar and the timeline for one of the days that it fails to show correctly

Odrakir:
The bug is hard to reproduce consistently. It happens often, but I'm not sure when.

I'm attaching four screenshots of the same month with different errors, only the first one is correct.

I entered the calendar, selected another day, went back to the calendar, changed month, selected another day... It ends up happening.

img_4389
img_4386
img_4387
img_4388

Odrakir:
And also, as I mentioned previously, the calendar is really slow (at least in debug mode). I'm afraid we'll have to dig deeper to fix this.

PR #974, opened 2018-01-24
🐛🚧Handle notification permisions rejection

### LINKED ISSUE #973: Notification issue
1. Disable notifications manually from your phone settings.
2. Open the app on the second day.
3. App asks to enable notifications. Pick 'No, thanks':
img_4971

4. Navigate to 'Goals'
5. Navigate back to 'Timeline'
6. This appears, again (being a bit broken):
img_4972
*After a while, 'No, thanks' option appears again and when I select it, the overlay closes. Though, when I repeat step 4, the overlay appears again in Timeline view.*

- For me, it happens only when I disable notifications manually from the phone settings
- This happens with current app store version and the latest build as well
- I'm using SE and iOS 11.2.1

sofokli1:
Closes #973

A flag in settings was not being set when selecting `No, thanks`


Odrakir:
Is this OK with Heidi's latest comment on the issue?

sofokli1:
the last commit should address the pop up showing only once after goal creation

sofokli1:
I have been trying many things with this but the solution still eludes me. The interesting part is that when the buttons show without text and you wait for some seconds the text appears. Fascinating issue. I am open to ideas for this. If no idea comes to mind I propose to merge this into develop but keep the issue open.

PR #997, opened 2018-02-01
Replace JTAppleCalendar by a custom implementation

### LINKED ISSUE #993: Replace current calendar by current solution
Current one is having problems and it's hard to migrate to Swift 4.

Odrakir:
Closes #993 

I simplified animations a bit for the moment, but otherwise, this works.

The way we get the activities will be improved in the future (we need interactors)


sofokli1:
I find some gliches:
- There is a white view that is being shown before the calendar and is removed after the calendar is removed
feb-05-2018 16-56-55

- There is a glitch on the month name that happens when using the arrows
feb-05-2018 16-58-51
- Also the day numbers are not aligned with the day labels on the top (especially visible on the Mondays and Sundays)
- The animation of the current month when opening the calendar seems a bit fast, but maybe that is just me

sofokli1:
When first opening the calendar the button to go to the next month is enabled and if pushed leads to crash (ends in the fatal error in the calendar view). If I scroll to the previews month and back to the current the button is disabled.

teferi/UI/Modules/Calendar/Views/CalendarView.swift
```diff
@@ -0,0 +1,171 @@
+import UIKit
+
+protocol CalendarDelegate
+{
+    func didChange(month: Date)
+    func didSelect(day: Date)
```
sofokli1: Why not RX instead of this protocol?

Odrakir: I just didn't feel it was worth it for to add that dependency for something so simple.

Odrakir:
@sofokli1 Please review now.

sofokli1:
If i open the calendar and select a date from the past month and reopen the calendar, the month name and year is wrong + the left arrow looks enabled even though it is the same month as the install date + both arrow do no work at all (the arrows and the month name and year work correctly if i scroll by hand to the next month and back)
feb-06-2018 11-50-51

Odrakir:
@sofokli1 Fixed, look again, please.

sofokli1:
Nothing has changed

PR #1006, opened 2018-02-06
Implement GetTimeSlotsForDate Interactor

### LINKED ISSUE #980: Add Interactor to retrieve TimeSlots
We'll introduce an InteractorFactory to produce Interactors for different use cases. Starting with this one.

This will fetch data from the different sources (motion, location and timeslots tables) to generate TimeSlot structs suitable for the business logic and to pass to the view models.

Odrakir:
Closes #980 

It has a lot of changes in the `TimelineViewModel` as well.


Odrakir:
This has been split into two PRs

PR #1009, opened 2018-02-07
Add `create` functionality to `CoreDataPersistency`

sofokli1:
Adds `create` functionality to `CoreDataPersistency`
and implements the `encode` func in the models


sofokli1:
@Odrakir I was thinking that it would make sense have a `create` method that accepts an array of models as well. What do you think?

Odrakir:
@sofokli1 Not for now, when we need to save multiple slots, it'll be the interactor looping through them... I think. Let's see how that works.

PR #1010, opened 2018-02-08
Get timeslots interactor

### LINKED ISSUE #980: Add Interactor to retrieve TimeSlots

Odrakir:
Closes #980


teferi/Services/Implementations/New Persistency/CoreDataPersistency.swift
```diff
@@ -23,6 +23,7 @@ class CoreDataPersistency
         return Observable.create {
             [unowned self] observer in
             
+            print("Perform fetch <-----------")
```
sofokli1: Remove print

Odrakir: Being this a cold observable is easy to subscribe twice to it and fire the fetch twice, that's why I'm printing... But yeah, let's find a better way to improve this. I'll remove it.

PR #1016, opened 2018-02-09
Core motion script

### LINKED ISSUE #978: Script to populate Motion Event database
The first time the user runs the app with the new algo, we have to get the CM data for the last week and populate with it the motion event table.

sofokli1:
Closes #978 

Adds an interactor to import motion events from the coreMotion to coreData. it is initialised in the appDelegate


sofokli1:
@Odrakir Let me know if the rx part in the interactor can be improved

teferi/Providers/Interactors/ImportMotionEvents.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+    
+    func execute() -> Observable<[MotionEvent]>
+    {
+        return appLifecycleService.movedToForegroundObservable
+            .flatMap(fetchLastMotionEvent)
+            .catchError { [unowned self] (_) -> Observable<MotionEventPM> in
+                return Observable.of(MotionEventPM(startTime: self.timeService.now.add(days: -7), motionType: MotionEventType.other))
+            }
+            .map({ $0.startTime.addingTimeInterval(1) })
```
Odrakir: Is this line needed? You are later on filtering them...

teferi/Providers/Interactors/ImportMotionEvents.swift
```diff
@@ -0,0 +1,55 @@
...
+            .flatMap(getMotionEvents)
+            .map(saveAsPM)
```
Odrakir: Saving, being a side effect, shouldn't be done on a `map`, a `flatMapLatest`is better. And it should return its result or `Observable<Void>`

teferi/Providers/Interactors/ImportMotionEvents.swift
```diff
@@ -0,0 +1,55 @@
...
+            return events.filter({ $0.start > since })
+        })
+    }
+    
+    private func saveAsPM(_ motionEvents: [MotionEvent]) -> [MotionEvent]
+    {
+        motionEvents
+            .map({ MotionEventPM(startTime: $0.start, motionType: $0.type) })
+            .forEach { (eventPM) in
+                persistency.create(eventPM).subscribe().disposed(by: disposeBag)
```
Odrakir: This is weird... I'm already not sure this interactor is a good idea by itself (maybe it's better to split it... not really sure). But anyway, here's how I would do it. You have to fill the methods than return `Observable.empty()`

Also `saveMotionEvent` will be part of the Repository. The idea is PM objects never get to the interactors, we'll move it when that part is ready.

```
    func execute() -> Observable<[MotionEvent]>
    {
        return getLastMotionDate()
            .flatMapLatest(getMotionEvents)
            .flatMapLatest(saveMotionEvents)
    }
    
    private func getLastMotionDate() -> Observable<Date>
    {
        return Observable.empty()
    }
    
    private func saveMotionEvent(motionEvent: MotionEvent) -> Observable<MotionEvent>
    {
        return Observable.empty()
    }
    
    private func saveMotionEvents(_ motionEvents: [MotionEvent]) -> Observable<[MotionEvent]>
    {
        return Observable.from(motionEvents)
            .flatMap(saveMotionEvent)
            .toArray()
    }
```

Podfile.lock
```diff
@@ -77,4 +77,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: bceeb49acea5c114b766d590220ff6b2f317cba5
 
-COCOAPODS: 1.4.0
+COCOAPODS: 1.1.1
```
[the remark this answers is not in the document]
sofokli1: will update

teferi/Providers/Interactors/ImportMotionEvents.swift
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return persistency.fetch(MotionEventPM.last())
+            .catchError { [unowned self] (error) -> Observable<MotionEventPM> in
+                if let persistencyError = error as? PersistencyError, persistencyError == .noResults
+                {
+                    return Observable.of(MotionEventPM(startTime: self.timeService.now.add(days: -7), motionType: MotionEventType.other))
+                } else {
+                    throw error
+                }
+            }.map({ $0.startTime })
```
Odrakir: If you put the `map` before the `catchError`you'll get the error anyway and you won't have to create that dummy `MotionEventPM`, you can just return a `Date`.

teferi/AppDelegate.swift
```diff
@@ -172,9 +172,15 @@ class AppDelegate : UIResponder, UIApplicationDelegate
             coreDataPersistency: CoreDataPersistency(managedObjectContext: coreDataStack.managedObjectContext),
             timeService: timeService,
             timeSlotService: timeSlotService,
-            appLifecycleService: appLifecycleService
+            appLifecycleService: appLifecycleService,
+            motionService: motionService
         )
         
+        InteractorFactory.shared.createImportMotionEventsInteractor()
```
Odrakir: I'm not really sure we want to do this just yet, as we are not using it for the moment. But in any case... it should probably run on a background thread, shouldn't it?

sofokli1: Since we are not using this now I would say better remove it. As per background threads I would wait until the current setup is a bit more robust and not changing much before optimising in this direction.

PR #1018, opened 2018-02-12
Abstract persistence

Odrakir:
This abstracts core data persistence and prepares for the next PR which introduces the Repository.


teferi/Services/Implementations/New Persistency/CoreDataResource.swift
```diff
@@ -1,46 +1,50 @@
...
-                throw PersistencyError.couldntParse
-            }
-            return try B.init(managedObject: first)
+        return CoreDataResource<B>(
+            entityName: B.entityName,
+            predicate: predicate == nil ? nil : NSPredicate(format: predicate!.format, argumentArray: predicate!.parameters),
+            sortDescriptors: sortDescriptor == nil ? nil : [sortDescriptor!],
+            parser: { managedObjects in
+                guard let first = managedObjects.first else {
+                    throw PersistencyError.couldntParse
```
sofokli1: Wouldn't it make sense to return nil instead of error when `managedObjects` is empty?

Odrakir: Yes, actually I think I'm probably going to return an array for both of them because it helps composition. But let me do that change in the repository PR.

PR #1021, opened 2018-02-12
Repository

Odrakir:
This adds a repository between the Interactors and DB to make access easier and through an abstraction.


teferi/Services/Implementations/New Persistency/CoreData/CoreDataResource.swift
```diff
@@ -42,17 +42,17 @@ extension CoreDataResource
...
+        return CoreDataResource<[B]>(
             entityName: B.entityName,
             predicate: predicate == nil ? nil : NSPredicate(format: predicate!.format, argumentArray: predicate!.parameters),
             sortDescriptors: sortDescriptor == nil ? nil : [sortDescriptor!],
             parser: { managedObjects in
                 guard let first = managedObjects.first else {
                     throw PersistenceError.noResults
                 }
-                return try B.init(managedObject: first)
+                return try [B.init(managedObject: first)]
```
sofokli1: shouldn't this be an empty array in case of no data?

teferi/Services/Implementations/New Persistency/CoreData/CoreDataPersistence.swift
```diff
@@ -45,13 +45,15 @@ class CoreDataPersistence: Persistence
         }
     }
     
-    func create<A: CoreDataModel>(_ model: A) -> Observable<A>
+    func create<A>(_ model: A) -> Observable<A>
     {
+        let coreDataModel = model as! CoreDataModel
```
sofokli1: Why have this here instead of in the func signature (as it was)?

teferi/Services/Implementations/New Persistency/RepositoryService.swift
```diff
@@ -0,0 +1,155 @@
...
+import RxSwift
+
+protocol RepositoryServiceProtocol
+{
+    func getTimeSlots(forDay date: Date) -> Observable<[TimeSlotEntity]>
+    func getLastMotionDate() -> Observable<Date?>
+    func saveMotionEvents(motionEvents: [MotionEvent]) -> Observable<Void>
+}
+
+class RepositoryService
```
sofokli1: Shouldn't this be more like `class RepositoryService : RepositoryServiceProtocol`?

teferi/Models/Persistence/MotionEventPM.swift
```diff
@@ -35,20 +35,32 @@ extension MotionEventPM: CoreDataModel
...
         let predicate = Predicate(parameter: "startTime", rangesFromDate: startDate as NSDate, toDate: endDate as NSDate)
         let sortDescriptor = NSSortDescriptor(key: "startTime", ascending: true)
         return CoreDataResource<[MotionEventPM]>.many(predicate: predicate, sortDescriptor: sortDescriptor)
     }
     
-    static func last() -> CoreDataResource<MotionEventPM>
+    static func last(before date: Date) -> CoreDataResource<[MotionEventPM]>
+    {
+        let predicate = Predicate(parameter: "startTime", isBefore: date as NSDate)
+        let sortDescriptor = NSSortDescriptor(key: "startTime", ascending: true)
```
sofokli1: shouldn't this be `ascending: false`?

teferi/Services/Implementations/New Persistency/CoreData/CoreDataResource.swift
```diff
@@ -42,17 +42,17 @@ extension CoreDataResource
         )
     }
     
-    static func single<B: CoreDataModel>(predicate: Predicate? = nil, sortDescriptor: NSSortDescriptor? = nil) -> CoreDataResource<B>
+    static func single<B: CoreDataModel>(predicate: Predicate? = nil, sortDescriptor: NSSortDescriptor? = nil) -> CoreDataResource<[B]>
```
sofokli1: Now the only difference between these 2 funcs is the name, because they get same parameters and return the same kine of data. Would it make sense to merge them into 2 and have a optional parameter for the number of items to return, if nil returns [allThatMachPredicate] else returns [topNElelmentsThatMachPredicate]?

teferi/Services/Implementations/New Persistency/CoreData/CoreDataPersistence.swift
Odrakir: Just in hopes of making the signature generic and independent from CoreData, as it comes from the protocol. It probably can be improved, but it's good enough for now.

teferi/Services/Implementations/New Persistency/CoreData/CoreDataResource.swift
Odrakir: Mmm... I see your point, and though I agree this looks a little inefficient, I don't want to litter the API with the "number of items to return" param. Let's leave it like this for now.

PR #1034, opened 2018-02-13
Import location from log file

### LINKED ISSUE #979: Get locations from log
The first one the user runs the app with the new algo, we should get locations from the log file and store them in the DB (the current code is broken and we are not storing them)

sofokli1:
Closes #979


teferi/Interactors/ImportLocations.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+    
+    func execute() -> Observable<Void>
+    {
+        guard let logURL = logURL, let data = try? String(contentsOf: logURL) else { return Observable.just(()) }
+        
+        let locations = splitToRows(data)
+                        .filter(isLocationRow)
+                        .map(tokenize)
+                        .map(cleanupTokens)
```
Odrakir: I think having two separate steps for `cleanupTokens` and `toLocationEntities` is a little unnecessary. I would do it in just one step may be using local vars to extract the tokens.

teferi/Interactors/ImportLocations.swift
```diff
@@ -0,0 +1,81 @@
...
+        formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+        return formatter
+    }()
+    
+    init(repositoryService: RepositoryService)
+    {
+        self.repositoryService = repositoryService
+    }
+    
+    func execute() -> Observable<Void>
```
Odrakir: I think we must add some kind of mechanism for this to return if this was already done before. Maybe checking dates in DB, using the preferences or just deleting the log file (we can change the name for the new one)

sofokli1: I was thinking since there will be a few of these interactors to execute only once (the first time the user updates the app), it would make sense to have a setting to check if migrated to new system and execute in appDelegate all of them. What do you think?

Odrakir: Sounds good 👍

sofokli1:
yes I did and we will see when it gets to the app store :trollface:

Issue #1039, opened 2018-02-14
Use notifications to notify slot change or new slot


Odrakir:
As we are moving away from TimeSlotService we won't be using `timeSlotsUpdatedObservable` and `timeSlotCreatedObservable` any more. We should instead use notifications fired from the appropiate interactors.

PR #1040, opened 2018-02-14
Edit times interactor

### LINKED ISSUE #1029: Add interactor to edit start times
Use it in EditTimesViewModel.

This logic has to change a little now that we only store start times

Odrakir:
Closes #1029 

Adds an interactor to edit times and adds also the repository and persistence methods for it.

This breaks the timeline update when editing times. That will get fixed in #1039


teferi/Models/Persistence/LocationEventPM.swift
```diff
@@ -31,42 +31,37 @@ extension LocationEventPM: CoreDataModel
...
         managedObject.setValue(accuracy, forKey: "horizontalAccuracy")
-        
-        return managedObject
     }
 }
 
 extension LocationEventPM
 {
-    static func all(fromDate: Date, toDate: Date) -> CoreDataResource<[LocationEventPM]>
+    static func all(fromDate: Date, toDate: Date) -> CoreDataResource<LocationEventPM>
```
sofokli1: Can you clarify this change? Shouldn't the result be an array?

Odrakir: In order to make different fetches composable, all of them should return arrays, that's why now we don't need to specify array in the Resource, because it'll always return and array of whatever type.

PR #1046, opened 2018-02-14
✨Creat timeSlot Intereactor

### LINKED ISSUE #1030: Add interactor to add a new TimeSlot
Use it from MainViewModel

sofokli1:
Closes #1030

It will be used after migration


teferi/Interactors/CreateTimeSlotWithCategory.swift
```diff
@@ -0,0 +1,22 @@
...
+    {
+        self.repositoryService = repositoryService
+        self.timeService = timeService
+        self.category = category
+    }
+    
+    func execute() -> Observable<Void>
+    {
+        let timeSlotEntity = TimeSlotEntity(startTime: timeService.now, endTime: nil, category: category, activities: [], edited: true)
+        return repositoryService.saveTimeSlotEntity(timeSlotEntity: timeSlotEntity)
```
Odrakir: I'd say it makes more sense to just send `startTime`, `category` and `createdByUser` instead of a `TimeSlotEntity`

teferi/Interactors/CreateTimeSlotWithCategory.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import RxSwift
+
+class CreateTimeSlotWithCategory: Interactor
```
Odrakir: I think we need to differentiate between two ways of creating TS. One is manually by the user and the other one is via smart guessing. How about we call this `ManuallyCreateTimeSlot` and it always sets the `edited` property to true?

Then later on we can create the interactor we need to create TSs from smart guesses.

teferi/Services/Implementations/New Persistency/RepositoryService.swift
```diff
@@ -45,13 +51,24 @@ class RepositoryService: RepositoryServiceProtocol
...
     }
+    
+    func saveTimeSlotEntity(timeSlotEntity: TimeSlotEntity) -> Observable<Void>
+    {
+        return Observable.of(timeSlotEntity)
+            .flatMap { [unowned self] timeSlot -> Observable<TimeSlotPM> in
+                let eventPM = TimeSlotPM(startTime: timeSlot.startTime, category: timeSlot.category, editedByUser: timeSlot.edited)
+                return self.persistence.create(eventPM)
+            }
+            .toArray()
```
Odrakir: `toArray` is not needed when just creating one entity. Actually, this whole method should be replaced by just what's inside of the `flatMap`... Let me know if you need an explanation on what we are doing for the other save methods.

teferi/Interactors/CreateTimeSlotWithCategory.swift
sofokli1: 👍

teferi/Interactors/CreateTimeSlotWithCategory.swift
sofokli1: 👍

teferi/Services/Implementations/New Persistency/RepositoryService.swift
sofokli1: I get it. It was late when I committed this 😄

teferi/Services/Implementations/New Persistency/RepositoryService.swift
```diff
@@ -69,14 +69,13 @@ class RepositoryService: RepositoryServiceProtocol
             .mapTo(())
     }
     
-    func saveTimeSlotEntity(timeSlotEntity: TimeSlotEntity) -> Observable<Void>
+    func saveManualTimeSlot(startTime: Date, category: Category, edited: Bool) -> Observable<Void>
```
Odrakir: At this point I don't need to know if the timeslot was created manually or not. I might reuse this method for smart guesses. So a better name would be `saveTimeSlot`

teferi/Services/Implementations/New Persistency/RepositoryService.swift
```diff
@@ -69,14 +69,13 @@ class RepositoryService: RepositoryServiceProtocol
...
     {
-        return Observable.of(timeSlotEntity)
-            .flatMap { [unowned self] timeSlot -> Observable<TimeSlotPM> in
-                let eventPM = TimeSlotPM(startTime: timeSlot.startTime, category: timeSlot.category, editedByUser: timeSlot.edited)
+        return Observable.of((startTime: startTime, category: category, edited: edited))
```
Odrakir: This is wrong is not needed here `Observable.of`, you are wrapping values into an Observable just to flatMap and return a new Observable. Totally unnecessary.

sofokli1: now that i am looking at it it looks, that looks awful

Odrakir: Oh, did you write it without looking at it? :trollface:

PR #1050, opened 2018-02-15
Edit category interactor

Odrakir:
Adds an interactor to edit the category and uses it from the timeline.

Right now the timeline itself doesn't update after that because it's missing the notifications, but that comes in another PR. To see the update just go to the previous day and back.


teferi/Interactors/EditTimeSlotCategory.swift
```diff
@@ -0,0 +1,23 @@
...
+import RxSwift
+
+class EditTimeSlotCategory: Interactor
+{
+    let repositoryService: RepositoryService
+    
+    let timeSlotsDates: [Date]
+    let category: Category
+    
+    init (repositoryService: RepositoryService, timeSlotsAt: [Date], category: Category)
```
sofokli1: why not just give the timeSlotEntities instead of dates? this implies that the inner circles knows how we handle timeSlotPMs in the outer circle

Odrakir: Mmmm... I'm not sure I agree, start times here add as an ID for timeslots, so it's the only way to uniquely identify them. To actually send a TimeSlot there I would have to build it or get it somehow first.

Maybe it's something we can do later on, in order to simplify the API, but right now we are mixing TimeSlot and TimeSlotEntity, but the Date stays the same.

Odrakir:
But the startTime IS the id, that's how we are finally making the edit in the DB (and has been like this since the start).

You are right that we should check e are not overwriting any other TS. That check is now being done in the UI, but the DB should do it as well. Let's create an issue for that.

sofokli1:
Have you ever had a project where changing the property of an object can change the ID of that object? (apart from this one 😄 ) Just because we have been doing it like that from the beginning doesn't make it right. I am not saying we should change what is ID in this PR, I am just putting it out there to start a conversation.

Odrakir:
Yeah, I'm fine with discussing it. And I think this behavior is very common. Business rules state you can't have two timeslots with the same startTime. It's the same as using the user's email as ID and at the same time allowing them to change it.

sofokli1:
Users can change their email because that is not their ID in the database, it is just a column like password, name, lastLoginTime etc.

Odrakir:
We are talking different things here. `startTime` isn't actually the real ID in the DB either. But I'm not referring to the specific DB implementation.

PR #1051, opened 2018-02-15
Custom notifications

Odrakir:
This uses NotificationCenter to communicate timeline changes between modules.

I think it only misses the notification posting on TS creation (and the test that checks that)


teferi/Interactors/EditTimeSlotStartTime.swift
```diff
@@ -17,6 +17,10 @@ class EditTimeSlotStartTime: Interactor
     
     func execute() -> Observable<Void>
     {
+        let date = originalDate
         return repositoryService.editSlotStartTime(from: originalDate, to: finalDate)
+            .do(onNext: { [unowned self] in
```
sofokli1: You are not using self in here. This gives a warning. Consider removing it

PR #1052, opened 2018-02-15
Fix timeslots not getting cropped to date

Odrakir:
Somehow we deleted the cropping, this brings it back.


teferi/Services/Implementations/New Persistency/RepositoryService.swift
```diff
@@ -23,6 +23,7 @@ class RepositoryService: RepositoryServiceProtocol
     {
         return persistence.timeSlots(forDay: date)
             .flatMapLatest(addActivities)
+            .map(cropToDay(date) >>> filterDurationZero)
```
sofokli1: I had a crash in `cropToDay` in line 177 `endTime` can be nil
`Fatal error: Unexpectedly found nil while unwrapping an Optional value`

Odrakir: I'm almost sure I fixed this in another PR, but let me fix it here too, just in case.

sofokli1: fix and merge the other PR and this should be fixed in the update of the branch

PR #1058, opened 2018-02-16
🚧Use interactor to create manual timeslot with the + button

sofokli1:
Doesn't close any issue


teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -222,15 +222,13 @@ class MainViewModel : RxViewModel
...
-                                             categoryWasSetByUser: true,
-                                             tryUsingLatestLocation: true)
-        else { return }
-        
-        metricsService.log(event: .timeSlotManualCreation(date: timeService.now, category: category))
-        metricsService.log(event: .timeSlotCreated(date: timeService.now, category: category, duration: nil))
+        InteractorFactory.shared.createManuallyCreateTimeSlottWithCategoryInteractor(category: category)
+            .execute()
+            .subscribe(onNext: { [unowned self] _ in
+                self.metricsService.log(event: .timeSlotManualCreation(date: self.timeService.now, category: category))
```
Odrakir: I think these two should be in the interactor as well.

teferi/UI/Modules/Main/MainViewModel.swift
```diff
@@ -222,15 +222,13 @@ class MainViewModel : RxViewModel
...
-                                             category: category,
-                                             categoryWasSetByUser: true,
-                                             tryUsingLatestLocation: true)
-        else { return }
-        
-        metricsService.log(event: .timeSlotManualCreation(date: timeService.now, category: category))
-        metricsService.log(event: .timeSlotCreated(date: timeService.now, category: category, duration: nil))
+        InteractorFactory.shared.createManuallyCreateTimeSlottWithCategoryInteractor(category: category)
+            .execute()
+            .subscribe(onNext: { [unowned self] _ in
```
Odrakir: I'm not a big fan of subscribing in the ViewModel, I rather return an Observable from addNewSlot and subscribe from ViewController to do something after success... But let's leave it like this for now.

teferi/Interactors/ManuallyCreateTimeSlot.swift
```diff
@@ -6,16 +6,21 @@ class ManuallyCreateTimeSlot: Interactor
...
     {
         self.repositoryService = repositoryService
         self.timeService = timeService
+        self.metricsService = metricsService
         self.category = category
     }
     
     func execute() -> Observable<Void>
     {
+        self.metricsService.log(event: .timeSlotManualCreation(date: self.timeService.now, category: self.category))
```
Odrakir: What happens if the creation fails for some reason? Then we probably don't want to send metrics. This two lines should be in a side effect to the `Observable`. (`do(onNext: {...}`)

sofokli1: do you mean like this?
```
    {
        return repositoryService.saveTimeSlot(startTime: timeService.now, category: category, edited: true)
            .do(onNext: { [unowned self] in
                self.metricsService.log(event: .timeSlotManualCreation(date: self.timeService.now, category: self.category))
                self.metricsService.log(event: .timeSlotCreated(date: self.timeService.now, category: self.category, duration: nil))
            })
    }
```

Odrakir: Yes, you could even extract them to a private method, but that's the idea.

sofokli1: I tried it like this before but had some issues with the `[unowned self]`

PR #1062, opened 2018-02-19
Replace `TimeSlot` with `TimeSlotEntity`

sofokli1:
Doesn't close any issue. Just replaces `TimeSlot` with `TimeSlotEntity`.


teferi/Extensions/ArrayExtensions.swift
```diff
@@ -117,6 +117,26 @@ extension Array where Element == TimeSlot
     }
 }
 
+extension Array where Element == MotionEventEntity
```
Odrakir: This code shouldn't be here. I even doubt it should be an array extension. Extensions should be used for more generic stuff, you shouldn't be calling this method from many places (in theory just once). I suggest it as a lazy property (or calculated on construction) on `TimeSlotEntity`, or on `SlotTimeLineItem`, depending on the usage, for example.

teferi/Extensions/ArrayExtensions.swift
```diff
@@ -117,6 +117,26 @@ extension Array where Element == TimeSlot
...
+{
+    func longest(withEndTime lastTime: Date?) -> Element?
+    {
+        guard count > 1 else { return self.first }
+        
+        var previousTime = lastTime
+        
+        return self.reversed()
```
Odrakir: This should calculate the longest activity type, not the longest single activity. Meaning, if there are several walk activities, the sum of all them, vs the sum of all still...

teferiTests/Helpers/SlotTimelineItemTestHelper.swift
```diff
@@ -3,24 +3,24 @@ import Foundation
...
     
-    static func with(timeSlot: TimeSlot, timeSlotService: TimeSlotService) -> SlotTimelineItem
+    static func with(timeSlotEntity: TimeSlotEntity, timeSlotService: TimeSlotService) -> SlotTimelineItem
     {
-        return SlotTimelineItem(withTimeSlots: [timeSlot],
-                            category: timeSlot.category,
-                            duration: timeSlotService.calculateDuration(ofTimeSlot: timeSlot),
+        return SlotTimelineItem(withTimeSlotEntities: [timeSlotEntity],
+                            category: timeSlotEntity.category,
+                            duration: timeSlotEntity.duration ?? 0,
```
Odrakir: How about using `timeService` to calculate duration here? If duration is nil, it's because its the last timeslot (the running one)

teferi/UI/Modules/Timeline/Views/SlotTimelineItem+TimelineCell.swift
```diff
@@ -102,4 +105,27 @@ extension SlotTimelineItem
             }
         }
     }
+    
+    private func longestMotionEvent(fromMotionEvents motionEvents: [MotionEventEntity], withEndTime lastTime: Date?) -> MotionEventEntity?
```
Odrakir: Call me picky, but this method isn't strictly part of the display logic. In that sense, it shouldn't be here but in `SlotTimelineItem`. `activityTagText` belongs here though.

Also, you are calling this method everytime someone asks for the `activityTagText`, which is potentially very inefficient.

Odrakir: But anyway, thinking about it... It might be better to just discard this approach completely. As before we could have many timeslots and now we have many timeslots with many activities. I think it's better if we just flatMap to the activities. 

Meaning: `firstActivity = timeSlotEntities.flatMap{ $0.activities }.first` in the previous code, and delete the new method alltogether.

What do you think?

sofokli1: we could go even simpler `firstActivity = timeSlotEntities.first?.activities.first?.motion`

Odrakir: Yeah, but you are going to need the activities for the reduce, right?

So:

```swift
let activities = timeSlotEntities.flatMap{ $0.activities }
guard let firstActivity = activities.first ...
let groupActivity = activities.dropFirst().reduce(...
```

PR #1069, opened 2018-02-20
Cleanup TimeSlotService usages

Odrakir:
This just deletes references to TimeSlotService where not needed anymore.

The stuff I'm deleting in the DetailViewModel is not used anymore. There's another issue to replace that with interactors.


teferi/Services/Protocols/Persistency/TimeSlotService.swift
```diff
@@ -33,9 +30,7 @@ protocol TimeSlotService
      - Parameter setByUser: Indicates if the user initiated the action that changed the TimeSlot.
      */
     func update(timeSlots: [TimeSlot], withCategory category: Category)
```
sofokli1: Why not remove this as well?

Odrakir: Correct. Removed.

PR #1072, opened 2018-02-20
🚧Make map show locations from tomeSlot

### LINKED ISSUE #1066: Show locations in TimeSlotDetailView
With the change of entities and UI they disappear, we have to bring them back.

sofokli1:
closes #1066

Tries to get `locations` from `[MotionEventEntites]` of `timeSlots`
The logic should be sound but we do not have activities in the timeSlots thus no locations to show


teferi/UI/Modules/Timeline/TimeSlotDetail/TimeslotDetailViewController.swift
```diff
@@ -21,6 +21,18 @@ class TimeslotDetailViewController: UIViewController
     @IBOutlet weak var endTimeButton: UIButton!
     
     @IBOutlet weak var map: MKMapView!
+    fileprivate let annotationViewIdentifier = "point"
+    private var dataSource : [Annotation]!
```
Odrakir: I don't think you need a property for this, you can bind the observable in line 82 directly.

PR #1073, opened 2018-02-20
Detail view actions

### LINKED ISSUE #1049: Bring back edition to TimeSlot detail view
We need to add the UI to edit the category (once the interactor is working) and modify a bit the time editing view model.

Odrakir:
Closes #1049 

Creates a new interactor to edit times replacing the other one that didn't consider all cases.


Odrakir:
The edit view alignment is fixed. The problem was it was taking the reference point wrong. So it shouldn't work for the timeline either, but some magic numbers in `EditTimeSlotView` were hiding the mistake.

Odrakir:
Actually there's still a problem with this, the edit view is not disappearing when it should... 🤔

sofokli1:
Try editing the startTime of slot 2 times in a row without leaving the details view

PR #1074, opened 2018-02-21
Remove TimeSlotService from WeeklySummayViewModel

### LINKED ISSUE #1024: Use Interactor to get timeslots in WeeklySummaryViewModel
We probably need to create a new one, or modify the existing one.

sofokli1:
Closes #1024 

The service is still needed for the category provider.
I added a variable to store the `timeSlots` for the current week in `WeeklySummayViewModel` and use that to get the `timeSlots` to the delegate of the chart view.


sofokli1:
@Odrakir Some feedback on the code would help a lot since you are more familiar with the component

Odrakir:
Going for a different approach.

PR #1075, opened 2018-02-21
Large titles for timeline

### LINKED ISSUE #966: Large titles for timeline
Add large titles to the timeline view

juanlaube:
Closes #966 

* Make navigation bar translucent

* Set navigation bar to use large title

* Add custom title view to navigation bar

* Pending: Fix collapse when scrolling and don't show large title and collapsed title view at the same time


Odrakir:
Hey @juanlaube I see this is done through the NavigationController, so the change takes effect in all sections at the same time.

Could you try implement the rest of the sections in this PR as well, see what other problems we find?

PR #1078, opened 2018-02-21
🚧Set endTime to timeSlots

### LINKED ISSUE #1067: Last TimelineItem shows always 0 duration
This is because instead of calculating the duration with the `timeService` when `endTime` is nil we are just returning 0.

sofokli1:
Closes #1067

Easy review


teferi/UI/Modules/Timeline/TimelineViewModel.swift
```diff
@@ -129,13 +129,20 @@ class TimelineViewModel: RxViewModel
         let timelineItems = timeSlotEntities
             .splitBy { $0.category }
             .flatMap { groupedTimeSlotEntities -> [TimelineItem] in
-                if groupedTimeSlotEntities.count > 1 && areExpanded(groupedTimeSlotEntities)
+                
+                var slots = groupedTimeSlotEntities
+                if let last = slots.last, last.endTime == nil
                 {
-                    return expandedTimelineItems(fromTimeSlotEntites: groupedTimeSlotEntities)
+                    slots = Array(slots.dropLast()) + [TimeSlotEntity(startTime: last.startTime, endTime: self.timeService.now, category: last.category, activities: last.activities, edited: last.edited)]
```
Odrakir: This is not where you should put this code. You are running it for every group of timeslots. You should only check the last timeslot of the last group. So you have to move this before the `let timelineItems = timeSlotEntities` or after this whole chain (probably inside the `if let lastTimelineItem = timelineItems.last, case TimelineItem.slot(let slotTimelineItem) = lastTimelineItem`

Odrakir: You can use `last.with(endTime:...)`

Issue #1026, opened 2018-02-12
Add interactor to get categories


Odrakir:
Instead of the DefaultCategoryProvider

PR #1079, opened 2018-02-21
WIP Add interactor to get categories (#1026)

juanlaube:


teferi/Interactors/GetDefaultCategories.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    init(repositoryService: RepositoryService, but categoriesToFilter: [Category] = [])
+    {
+        self.repositoryService = repositoryService
+        self.categoriesToFilter = categoriesToFilter
+    }
+
+    func execute() -> Observable<[Category]>
+    {
+        let today = Date().ignoreDateComponents()
```
Odrakir: We use `TimeService` to get the current time, this is just so we can test easily later on.

teferi/Interactors/GetDefaultCategories.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import RxSwift
+
+class GetDefaultCategories: Interactor
+{
+
+    private let repositoryService: RepositoryService
+    private let categoriesToFilter: [Category]
+
+    init(repositoryService: RepositoryService, but categoriesToFilter: [Category] = [])
```
Odrakir: The `but` doesn't read very well... I would remove it.

teferi/Interactors/GetDefaultCategories.swift
juanlaube: Done!

teferi/Interactors/GetDefaultCategories.swift
juanlaube: Removed

juanlaube:
I've removed the category extension :)

teferi/Interactors/GetDefaultCategories.swift
```diff
@@ -0,0 +1,55 @@
...
+    {
+        let today = timeService.now.ignoreTimeComponents()
+
+        let startTime = today.add(days: -14).ignoreTimeComponents()
+        let endTime = today.tomorrow.ignoreTimeComponents()
+
+        return repositoryService
+            .getTimeSlots(fromDate: startTime, toDate: endTime)
+            .map({ timeslots in
+                self.allSorted(byUsage: timeslots).filter { !self.categoriesToFilter.contains($0) }
```
Odrakir: This code is fine as it is right now. But the whole idea behind this change is that we'll get the categories from the repositoryService as well. So let's try to change this so instead of using `Category.all`we get the categories with this method (which mocks what the repository will do):
```swift
    func allCategories() -> Observable<[Category]>
    {
        return Observable.just(Category.all)
    }
```

Then you'll have to sort by timeslot usage.

Also you can skip the filtering as you are doing it in the `CategoryProvider` already.

teferi/UI/Common/Views/EditTimeSlotView.swift
```diff
@@ -44,6 +44,13 @@ class EditTimeSlotView : UIView, TrigonometryHelper, CategoryButtonDelegate
     private var leftBoundryX : CGFloat!
     private var rightBoundryX : CGFloat!
     private var categoryProvider : CategoryProvider!
+    private var categoryProviderObservable : Observable<CategoryProvider>! {
+        didSet {
+            categoryProviderObservable.subscribe(onNext: { provider in
```
Odrakir: The problem with this approach is that it's hard to handle errors (once we do handle errors). I would rather remove this observable and remove the categoryProvider constructor injection leaving it as a public property for property injection. 

Then you can subscribe to the `categoryProviderObservable` from the VC (`MainViewController`, for example) and inject the `categoryProvider` in the subscription.

teferi/ViewModelLocator.swift
```diff
@@ -257,7 +257,7 @@ class DefaultViewModelLocator : ViewModelLocator
                                 goalService: goalService,
                                 notificationService: notificationService,
                                 settingsService: settingsService,
-                                categoryProvider: DefaultCategoryProvider(timeSlotService:  timeSlotService),
+                                categoryProvider: InteractorFactory.shared.createGetDefaultCategories().execute().map({ DefaultCategoryProvider(categories: $0) }),
```
Odrakir: I don't know why we are injecting the category provider here in the first place. Remove it from the constructor parameters.

teferi/UI/Modules/Goals/NewGoal/NewGoalViewModel.swift
```diff
@@ -7,17 +7,18 @@ class NewGoalViewModel
     private let goalService: GoalService
     private let notificationService: NotificationService
     private let settingsService: SettingsService
-    private let categoryProvider: CategoryProvider
+    private var categoryProvider: CategoryProvider! {
```
Odrakir: Actually, we don't need a categoryProvider here at all. Just with the `GetDefaultCategoriesSorted` interactor is enough. We can bind it directly to the categories observable (not Variable)

teferi/UI/Modules/Main/Views/AddTimeSlotView.swift
```diff
@@ -30,10 +30,17 @@ class AddTimeSlotView : UIView
             self.wheel.categoryProvider = categoryProvider
         }
     }
+    var categoryProviderObservable : Observable<CategoryProvider>! {
```
Odrakir: As mentioned before, same deal, better remove the observable inject the categoryProvider from outside.

teferi/Interactors/GetDefaultCategories.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+import RxSwift
+
+class GetDefaultCategories: Interactor
```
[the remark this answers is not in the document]
juanlaube: Renamed

teferi/UI/Common/Views/EditTimeSlotView.swift
juanlaube: Agreed, that's much more clean.

